PR #37, opened 2021-02-24
Update CLI README.md

Erin-Cecele:
Replace install / update/ uninstall with new instructions.
Add Quire Commands.

I believe "Developing Quire CLI" and "Testing Quire CLI" instructions are out-of-date and should updated or removed. Leaving them in place for now as a reminder.


geealbers:
Seeing this now and it being so verbose and such a close (exact?) copy of the information from our documentation, I really think this is going to only add a maintenance headache, as you're going to have to track that every time we make a change to the docs, it's going to have to be made it here too, and vice versa.

I'd recommend really stripping this down to at most being the Quick Install instructions and the table of the commands (which could both be a verbatim copy from the main README), and then the Developing and Testing sections, which in my mind are actually the unique things that really belong here (as they don't belong anywhere else) even if they are currently out of date for now.

geealbers (changes requested):
Seeing this now and it being so verbose and such a close (exact?) copy of the information from our documentation, I really think this is going to only add a maintenance headache, as you're going to have to track that every time we make a change to the docs, it's going to have to be made it here too, and vice versa.

I'd recommend really stripping this down to at most being the Quick Install instructions and the table of the commands (which could both be a verbatim copy from the main README), and then the Developing and Testing sections, which in my mind are actually the unique things that really belong here (as they don't belong anywhere else) even if they are currently out of date for now.

geealbers (approved):

PR #51, opened 2021-03-04
Update CONTRIBUTING.md for new repo WIP

Erin-Cecele:
A few questions/comments:

- I wasn't sure how best to describe the main code repository so I kept it simple
- Wasn't sure how to handle contributions to the documentation. Should issues for the docs be tracked in main repo or on docs repo? We'll need to point them to the docs repo at some point so I directed people to post issues there. Is that too confusing? Should all issues remain in one place?
- I made small edits to the section about new features/themes but please take a closer look at this.
- Removed mention of Github article (let's discuss on a call)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,31 +2,28 @@
 
 **Thank you for your interest in contributing to Quire!**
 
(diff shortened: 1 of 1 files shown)


geealbers (dismissed):
I think this all seems fine. I'm sure we'll adjust as we go and actually get people who want to contribute.

On the question of issue tracking for the documentation, I think what you're doing is right and that we should maintain those issues in that repo and direct people to both places rather than consolidate them all into the main repo.

PR #54, opened 2021-03-09
Refactor/DEV-8270: Remove file extension from iiif image output directory

hbalenda:
- Renames variables to more clearly indicate original vs destination paths
- Uses path.parse to derive filenames and extensions
- Adds `-noextension` option to go-iiif command to remove file extension from output directory name


mphstudios (approved):
@hbalenda the renaming here is good.

PR #67, opened 2021-03-16
Fix/DEV-7773: Contain citation popups within window

hbalenda:
Changes:
- Check the position of the citation, and if it's outside to the left or right of its container, move it within the container boundary.

### DIFF
--- a/themes/default/source/css/components/quire-citation.scss
+++ b/themes/default/source/css/components/quire-citation.scss
@@ -63,7 +63,7 @@
     z-index: 1;
     bottom: 150%;
     left: 50%;
-    margin-left: -200px;
+    transform: translateX(-200px);
     -webkit-box-shadow: 0px 0px 6px 0px $black-semi-transparent;
     -moz-box-shadow: 0px 0px 6px 0px $black-semi-transparent;
     box-shadow: 0px 0px 6px 0px $black-semi-transparent;
@@ -71,12 +71,12 @@
     @media screen and (max-width: $desktop) {
       width: auto;
       right: 0;
-      margin-left: -100px;
+      transform: translateX(-100px);
     }
     @media screen and (min-width: $desktop) and (max-width: 978px) {
       width: 200px;
       right: 0;
-      margin-left: -100px;
+      transform: translateX(-100px);
     }
   }
 }
--- a/themes/default/source/js/application.js
+++ b/themes/default/source/js/application.js
@@ -473,6 +473,30 @@ function validateSize(map) {
 }
 
 /**
+* Translates the X-position of an element inside a container so that its contents
+* are contained
+*
+* @param {object} element to position
+* @param {object} container element
+* @param {number} container margin
+*/
+function setPositionInContainer(el, container, margin = 0) {
+  const elRect = el.getBoundingClientRect();
+  const containerRect = container.getBoundingClientRect();
+
+  const leftDiff = elRect.left - containerRect.left;
+  const rightDiff = containerRect.right - elRect.right;
+  const halfElWidth = elRect.width/2;
+  // x
+  if (rightDiff < 0) {
+    el.style.transform = `translateX(-${halfElWidth-rightDiff+margin}px)`;
+  } else if (leftDiff < 0) {
+    el.style.transform = `translateX(-${halfElWidth+leftDiff+margin}px)`;
+  }
+  // @todo y
+}
+
+/**
  * @description
  * find expandable class and look for aria-expanded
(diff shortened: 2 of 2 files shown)


themes/default/source/js/application.js
```diff
@@ -472,6 +472,30 @@ function validateSize(map) {
...
 
+/**
+* Translates the X-position of an element inside a container so that its contents
+* are contained
+*
+* @param {object} element to position
+* @param {object} container element
+* @param {number} container margin
+*/
+function setPositionInContainer(el, container, margin = 0) {
```
mphstudios, line 483: Is `containElementWithinParent` a more explicit method name?

themes/default/source/js/application.js
```diff
@@ -472,6 +472,30 @@ function validateSize(map) {
...
+
+  const leftDiff = elRect.left - containerRect.left;
+  const rightDiff = containerRect.right - elRect.right;
+  const halfElWidth = elRect.width/2;
+  // x
+  if (rightDiff < 0) {
+    el.style.transform = `translateX(-${halfElWidth-rightDiff+margin}px)`;
+  } else if (leftDiff < 0) {
+    el.style.transform = `translateX(-${halfElWidth+leftDiff+margin}px)`;
+  }
```
mphstudios, line 495: @hbalenda do we need to add a comment on `halfElWidth` that it is used because the element css selector already translates the position to be centered above an item? Should the sign on the formula be `-` when translating the element position to the left and to the right? For instance:
```suggestion
  if (rightDiff < 0) {
    el.style.transform = `translateX(-${halfElWidth-rightDiff-margin}px)`;
  } else if (leftDiff < 0) {
    el.style.transform = `translateX(+${halfElWidth+leftDiff+margin}px)`;
  }
```

themes/default/source/js/application.js
```diff
@@ -472,6 +472,30 @@ function validateSize(map) {
...
+* @param {object} container element
+* @param {number} container margin
+*/
+function setPositionInContainer(el, container, margin = 0) {
+  const elRect = el.getBoundingClientRect();
+  const containerRect = container.getBoundingClientRect();
+
+  const leftDiff = elRect.left - containerRect.left;
+  const rightDiff = containerRect.right - elRect.right;
+  const halfElWidth = elRect.width/2;
```
mphstudios, line 489: Adding a comment on `halfElWidth` would be kind to our future selves; we are not likely to remember that the element selector already translates the position to be centered above an item so this must be half its width.

mphstudios (changes requested):
@hbalenda this looks good though double check the sign of the `translate` formula when the element needs to be shifted to the right to stay within the containing element.

themes/default/source/js/application.js
hbalenda: You're right, this was wrong. I changed the definitions of rightDiff and leftDiff to be positive if the element is out of bounds so it's easier to read and fixed the margin math.

mphstudios (approved):

themes/default/source/css/components/quire-citation.scss
```diff
@@ -63,20 +63,20 @@
...
       width: auto;
       right: 0;
-      margin-left: -100px;
+      transform: translateX(-100px);
     }
     @media screen and (min-width: $desktop) and (max-width: 978px) {
       width: 200px;
       right: 0;
-      margin-left: -100px;
+      transform: translateX(-100px);
```
geealbers, line 79: @hbalenda Having `width: auto` on line 72 doesn't work here, the citation box defaults to being very very skinny, at least on Firefox. 

My fix was to set that width to 200px, and then just delete lines 76–80. One break point is enough.

geealbers (commented):
@hbalenda I'm porting this fix to the Rococo catalogue now and ran into a couple issues. Wasn't sure where the best place to put these comments was though, on this closed PR like I'm doing, or in Jira, or somewhere else. Let me know for the future.

Along with the comment I made here about `width: auto` could you also adjust the translations a little bit so that the citations boxes aren't not flush with edge of the window as in this screenshot? Maybe just move them 6–8px inward? Thanks!

Screen Shot 2021-03-26 at 11 24 22 AM

themes/default/source/css/components/quire-citation.scss
hbalenda: @geealbers Yep, totally, I can make those changes today and will tag you. In the future, if the PR/task is closed, it's preferable to open a new JIRA ticket and assign to me so I can track the work.

PR #68, opened 2021-03-16
Fix/DEV-8262: Use path module for cross-OS support

hbalenda:
Changes:
- Replace hard-coded path separators with path module
- Replace `outputDir` with `this.config.publishDir`
- Add `runPandoc` method to handle pandoc commands in `buildEpub` and `buildMobi`.

### DIFF
--- a/packages/cli/lib/project.js
+++ b/packages/cli/lib/project.js
@@ -68,14 +68,14 @@ class Project extends EventEmitter {
   * @param {string} the destination for the output file
   * @param {string} the path to the cover image
   */
-  runPandoc(outputFile, cover) {
+  runPandoc(outputFile, coverImage) {
     // @see https://pandoc.org/MANUAL.html#general-options
     const args = [
       `--from=html-native_divs+native_spans`,
       `--to=epub ${path.join("html","epub.xhtml")}`,
       `--output=${outputFile}`,
       `--epub-metadata=${path.join("html", "dc.xml")}`,
-      `--epub-cover-image=${cover}`,
+      `--epub-cover-image=${coverImage}`,
       `--template=${path.join("html", "template.xhtml")}`,
       `--css=${path.join(this.config.publishDir, "css", "epub.css")}`,
       `--standalone`
@@ -527,13 +527,15 @@ class Project extends EventEmitter {
           const filePath = path.dirname(fileNamePath);
           const fileName = path.basename(fileNamePath);
           await fs.ensureDir(filePath);
-          let cover =
-            epub.data.cover !== undefined
-              ? epub.data.cover.replace("http://localhost:1313/img/", "")
-              : "";
-          cover = cover !== "" ? `${path.join("static", "img", cover)}` : "";
+          const coverImage = epub.data.cover
+            ? path.join(
+                "static",
+                "img",
+                epub.data.cover.replace("http://localhost:1313/img/", "")
+              )
+            : "";
           const outputFile = `${fileNamePath}.epub`;
-          this.runPandoc(outputFile, cover);
+          this.runPandoc(outputFile, coverImage);
           spinner.succeed(`Filepath: ${fileNamePath}.epub`);
           spinner.info(
             `Execution Time: ${((performance.now() - start) / 1000).toFixed(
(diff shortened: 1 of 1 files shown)


mphstudios (approved):
@hbalenda approved, though before merging, please search for other instances where the OS path delimiter may be hardcoded and add a correction to this PR.

geealbers:
@hbalenda I actually ran into an issue last week that **might** be related. I was trying to run `quire epub` on a Windows machine. The output failed and the error message I got seemed to suggest that PrinceXML wasn't finding the pages at the URL it thought they should be at. I didn't have a chance to further test or document the apparent bug, but I can look later today and at least post the error messaging if it might be useful and related to what you're fixing here. Otherwise I can file it as a separate issue later.

hbalenda:
@geealbers - there are quite a few places where the path delimiters are hard-coded which i'm working on changing now. It might be better to wait til that's done to test & create new issues.

packages/cli/lib/project.js
```diff
@@ -63,6 +63,29 @@ class Project extends EventEmitter {
...
+      `epub`,
+      path.join("html","epub.xhtml"),
+      `-o`,
+      outputFile,
+      `--epub-metadata=${path.join("html", "dc.xml")}`,
+      cover,
+      `--template=${path.join("html", "template.xhtml")}`,
+      `--css=${path.join(this.config.publishDir, "css", "epub.css")}`,
+      `-s`
+    ];
```
mphstudios, line 85: Using unabbreviated command flag-name=value makes this more readable and is more explicit, documents, the command being run.
```suggestion
    // @see https://pandoc.org/MANUAL.html#general-options
    const args = [
      '--from=html-native_divs+native_spans',
      `--to=epub ${path.join("html", "epub.xhtml")}`,
      `--output=${outputFile}`,
      `--epub-metadata=${path.join("html", "dc.xml")}`,
      cover,
      `--template=${path.join("html", "template.xhtml")}`,
      `--css=${path.join(this.config.publishDir, "css", "epub.css")}`,
      '--standalone'
    ];
```

packages/cli/lib/project.js
```diff
@@ -639,31 +648,20 @@ class Project extends EventEmitter {
...
           await fs.ensureDir(filePath);
           let cover =
             epub.data.cover !== undefined
               ? epub.data.cover.replace("http://localhost:1313/img/", "")
               : "";
-          cover = cover !== "" ? `--epub-cover-image=static/img/${cover}` : "";
-          const fileNamePath = path.join(filePath, fileName);
-          let args = `-f html-native_divs+native_spans -t epub html/epub.xhtml -o ${fileNamePath}-mobi.epub --epub-metadata=html/dc.xml ${cover} --template=html/template.xhtml --css=${outputDir}/css/epub.css -s`;
-          execSync(`pandoc ${args}`);
+          cover = cover !== "" ? `--epub-cover-image=${path.join("static", "img", cover)}` : "";
```
mphstudios, line 662: To be consistent with how the other command args are being handled thos should be the cover file path only, without the command arg flag; move `--epub-cover-image` to the command args array above.

packages/cli/lib/project.js
hbalenda: Great suggestion, this is way more readable

mphstudios (approved):

packages/cli/lib/project.js
```diff
@@ -639,31 +645,20 @@ class Project extends EventEmitter {
...
           await fs.ensureDir(filePath);
           let cover =
             epub.data.cover !== undefined
               ? epub.data.cover.replace("http://localhost:1313/img/", "")
               : "";
-          cover = cover !== "" ? `--epub-cover-image=static/img/${cover}` : "";
-          const fileNamePath = path.join(filePath, fileName);
-          let args = `-f html-native_divs+native_spans -t epub html/epub.xhtml -o ${fileNamePath}-mobi.epub --epub-metadata=html/dc.xml ${cover} --template=html/template.xhtml --css=${outputDir}/css/epub.css -s`;
-          execSync(`pandoc ${args}`);
+          cover = cover !== "" ? `${path.join("static", "img", cover)}` : "";
```
mphstudios, line 659: Do you think this expression is a bit odd, it reads "if the cover image **is not an _empty string_**, use the path to the cover otherwise **use an _empty string_**."?

mphstudios, line 659: Is `coverImage` a more explicit name for this variable?

mphstudios (approved):

PR #71, opened 2021-03-22
Fix/DEV-7105: Load image sets with only IIIF images

hbalenda:
Fixes: 
- Previously, `deepZoomSetup`  was a callback of `preloadImages` which is only called on images with a `src` attribute. Changing `preloadImages` to a function that returns a promise, and resolves when passed an empty array so that `deepZoomSetup` and `mapSetup` are called whether or not images with `src` attributes are present.

### DIFF
--- a/themes/default/source/js/helper.js
+++ b/themes/default/source/js/helper.js
@@ -107,19 +107,12 @@ export const enableScroll = element => {
  * @return {Promise}
  */
 export const preloadImages = (srcs) => {
-  return new Promise((resolve, reject) => {
-    if (!srcs.length) {
-      resolve();
-    }
-    let img;
-    let remaining = srcs.length;
-    for (let i = 0; i < srcs.length; i++) {
-      img = new Image();
+  return new Promise((resolve) => {
+    if (!srcs.length) resolve();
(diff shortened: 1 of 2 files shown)


themes/default/source/js/helper.js
```diff
@@ -102,22 +102,28 @@ export const enableScroll = element => {
...
+      img = new Image();
+      img.onload = () => {
+        --remaining;
+        if (remaining <= 0) {
+          resolve();
+        }
+      };
+      img.src = srcs[i];
+    }
+  });
```
mphstudios, line 126: Is there a situation for which we need to `reject` the Promise and `throw` an error? We might loop through the array backwards, like:
```suggestion
    if (!srcs.length) resolve();
    for (let i = srcs.length; i >= 0; --i) {
      const img = new Image();
      img.src = srcs[i];
      img.onload = () => {
        if (i <= 0) resolve();
      };
    }
```

mphstudios (changes requested):
@hbalenda this looks good though see inline comment on the `preloadImages` helper function.

mphstudios (approved):

PR #72, opened 2021-03-22
Fix/DEV-8259: Replace starter project .gitignore

hbalenda:
Moving to the monorepo we combined all the `.gitignore` files at the root, but new Quire projects still need a `.gitignore`.

### DIFF
--- a/starters/default/.gitignore
+++ b/starters/default/.gitignore
@@ -1,0 +1,16 @@
+# Quire starter project ignored files
+
+.DS_Store
+.bundle
+.cache
+.sass-cache
+public
+site
+node_modules
+.tmp
+build
+Gemfile.lock
+themes/default/static/css/application.css
+themes/default/static/js/application.js
+themes/default/tests/e2e/
+themes/default/test-results/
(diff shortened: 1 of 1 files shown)


starters/default/.gitignore
```diff
@@ -0,0 +1,16 @@
...
+public
+site
+node_modules
+.tmp
+build
+Gemfile.lock
+themes/default/static/css/application.css
+themes/default/static/js/application.js
+themes/default/tests/e2e/
+themes/default/test-results/
```
mphstudios, line 16:
```suggestion
# Quire starter project ignored files
.DS_Store
.bundle
.cache
.sass-cache
.tmp
build
node_modules
public
site
themes/default/static/css/application.css
themes/default/static/js/application.js
themes/default/test-results/
themes/default/tests/e2e/

```

mphstudios (changes requested):
`Gemfile.lock` is a legacy from a prior Ruby implementation of Quire.

mphstudios (approved):

PR #76, opened 2021-03-23
Refactor/DEV-8376: Rename quire imageslice -> quire process --iiif

hbalenda:
Changes:
- Rename `quire imageslice` -> `quire process --iiif`
- Log error if process option is not provided
- Log error if process option does not call a project function
- Add jsdocs


mphstudios (approved):
Nice!

PR #78, opened 2021-03-24
Fix/DEV-8383: Set zoom level from YAML

hbalenda:
Changes:
- Rename `figures.yaml` image configuration `zoom_level` to `zoom_max` to be more explicit
- Set default max number of zoom levels in Leaflet to 6
- Set max number of scale factors generated by the image tiler to 6 
- Set `data-zoom-max` to a figure's `zoom_max` property, if one is present in `figures.yaml`
- Use `data-zoom-max` attribute in `deepzoom.js` to override leaflet default max zoom

Changes apply to all zoomable images, not just IIIF images.

### DIFF
--- a/themes/default/source/js/deepzoom.js
+++ b/themes/default/source/js/deepzoom.js
@@ -28,8 +28,8 @@ class DeepZoom {
     this.iiif = this.el.getAttribute("data-iiif");
 
     const zoom = {
-      min: 1,
       default: 1,
+      min: 1,
       max: this.el.getAttribute('data-zoom-max') || 6
     };
 
--- a/themes/default/source/js/deepzoom.js
+++ b/themes/default/source/js/deepzoom.js
@@ -5,7 +5,8 @@ import "leaflet-iiif";
 class DeepZoom {
   // add map array to access map objects outside of Deepzoom class
   constructor(id, mapArr) {
-    this.el = id;
+    this.id = id;
+    this.el = document.querySelector(`#${this.id}`);
 
     // remove and refresh before init
     // @ts-ignore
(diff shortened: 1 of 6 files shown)


themes/default/source/js/deepzoom.js
```diff
@@ -22,8 +23,15 @@ class DeepZoom {
...
-    this.imageURL = $(`#${this.el}`).attr("src");
-    this.iiif = $(`#${this.el}`).data("iiif");
+    this.catalogueEntry = this.el.getAttribute("data-catalogue-entry");
+    this.imageURL = this.el.getAttribute("src");
+    this.iiif = this.el.getAttribute("data-iiif");
+
+    const zoom = {
+      min: 1,
+      default: 1,
+      max: this.el.getAttribute('data-zoom-max') || 6
```
mphstudios, line 33:
```suggestion
      default: 1,
      min: 1,
      max: this.el.getAttribute('data-zoom-max') || 6
```

mphstudios (approved):

mphstudios (approved):

hbalenda:
@geealbers  - I changed the figures.yaml image option from `zoom_level` to `zoom_max` to be more explicit, but that should work as expected now that this is merged.

geealbers:
Thanks @hbalenda! Just tested it out and it works for me as expected.

Just one gut check. 

Having the name reversed to `max_zoom` sounds more natural to me, I think because one would typically say out loud, "What's the maximum zoom?" rather than "What's the zoom maximum?"

However, I'm assuming from a code perspective that it's preferable to have the naming as `zoom_max`  because it implies that `max` is an property of `zoom` so if we were to have additional properties they'd group together nicely like `zoom_min`, `zoom_speed` or whatever. Is that right? (I ask because when I was first testing it out, I wasn't being careful and defaulted to using `max_zoom` and it took me longer than it should have to figure it out.)

hbalenda:
@geealbers Exactly, there might be other zoom properties (zoom_default, for example, I could definitely see being used) that we would want to configure and have grouped together. I don't mind changing it though, up to you.

geealbers:
@hbalenda cool. makes sense to leave it as is. thanks!

PR #79, opened 2021-03-24
Feature/DEV-8372: Use git-lfs in new projects

hbalenda:
Changes:
- Run `git-lfs` command in new projects to track anything in the `/img` directory
- Log warning if user doesn't have `git-lfs` installed
- Update Quire docs link in kindle previewer warning

The Quire docs will also need to be updated to instruct users to install `git-lfs`:  https://github.com/git-lfs/git-lfs/wiki/Installation

### DIFF
--- a/packages/cli/lib/cli.js
+++ b/packages/cli/lib/cli.js
@@ -101,7 +101,7 @@ export default class CLI extends EventEmitter {
         if (commandMissing("git-lfs")) {
           this.warn(`Warning: Please install "git-lfs". This is recommended for managing file storage, such as large images. Please see the Quire docs for instructions on how to install it: https://quire.netlify.app/`);
         } else {
-          spawnSync(`git-lfs track "img/**/*"`);
+          spawnSync(`git-lfs`, [`track`, `"img/**/*"`]);
         }
         this.notice("Committing starter files...");
         spawnSync("git", ["add", "-A"]);
--- a/packages/cli/lib/cli.js
+++ b/packages/cli/lib/cli.js
@@ -98,6 +98,12 @@ export default class CLI extends EventEmitter {
         this.notice("Initializing git in the new project directory...");
         process.chdir(projectDir);
         spawnSync("git", ["init"]);
+        const lfsCmd = `git-lfs track "img/**/*"`;
+        if (commandMissing("git-lfs")) {
+          this.warn(`Warning: Please install "git-lfs". This is recommended for managing file storage, such as large images. Please see the Quire docs for instructions on how to install it: https://quire.netlify.app/`);
+        } else {
+          spawnSync(`lfsCmd`);
+        }
         this.notice("Committing starter files...");
         spawnSync("git", ["add", "-A"]);
         spawnSync("git", ["commit", "-m", `Add starter and theme to project`]);
--- a/packages/cli/lib/cli.js
+++ b/packages/cli/lib/cli.js
@@ -98,11 +98,10 @@ export default class CLI extends EventEmitter {
         this.notice("Initializing git in the new project directory...");
         process.chdir(projectDir);
         spawnSync("git", ["init"]);
-        const lfsCmd = `git-lfs track "img/**/*"`;
         if (commandMissing("git-lfs")) {
           this.warn(`Warning: Please install "git-lfs". This is recommended for managing file storage, such as large images. Please see the Quire docs for instructions on how to install it: https://quire.netlify.app/`);
         } else {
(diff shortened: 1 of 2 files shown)


mphstudios (dismissed):

geealbers:
@hbalenda if we're going to move forward with this, we should track everything in the `/downlaods` directory as well. The PDF and e-book files can quickly get to GitHub's 100MB file limit.

However, before you merge this in, I'd like us to think it through a little bit more. @Erin-Cecele should weigh in on the wording of the warning messages, and I think we need to test this on a machine that doesn't have git-lfs already installed so that we fully understand the user experience. I'm a little concerned that we're adding a layer of complication to the process that solves a potential future problem only for a minority of Quire's users, most of whom we've found currently don't use a git-based workflow. (And I'm considering here that even the mental energy of reading and reacting to a warning message might be a complication to many users.) Though that said, we are encouraging users to move to GitHub, and so getting them set up on git-lfs from the get go might be a great service to them. I'm just not sure.

@Erin-Cecele, what do you think? And do you have git-lfs installed on your laptop? I think you might from having worked on Rococo, but you can check with `git lfs version`. These are the questions I'd be interested in answering, and perhaps you have more:

- How does the warning message read in the Terminal output, and how can we expect users to react?
- What is the experience of adding files to a Quire repo without git-lfs installed and then adding that repo to GitHub desktop later? I'm imaging some users might get further along with their projects and _then_ decide to use GitHub. What will their first commit look like? When will GH Desktop tell/warn them about git-lfs?
- What is the experience of installing git-lfs?
- Finally, after installing git-lfs what is the experience of then pushing their repo to github.com for the first time?

Erin-Cecele:
@geealbers @hbalenda I actually have not used Git LFS before. I'm happy to install it and give it a go to get a sense of what the user experience is like. Then I can speak to many of @geealbers's questions above. 

Taking a few steps back, I'd like to get a better understanding of at what point in the workflow process users would be receiving this warning message. That will give me a better sense of an effective message.

Questions:

- Is this message coming up anytime users are adding images to their project (regardless of having a GH repo or having large image sizes)?
- Are we suggesting that running git-lfs would be part of the process every time someone runs either `quire new` for a new project or `quire install` for a cloned project? 
- What is the larger problem this mandatory install of git-lfs is solving beyond simplifying processes in iiif?
- I have a feeling I know the answer to this, but is there any chance this could be optional? There are definitely users this added step would benefit but not the majority. 

In response to @geealbers's comments, while we are working to get more users up-and-running on GH, it will take some time to convert our user base who is, for the most part, very intimidated by GH. It will take training, some handholding, and a gradual normalizing of certain processes. It will not happen overnight. So, I am also hesitant to add another level of complication around using (and understanding) Quire, especially for the large percentage of users not on GitHub. It could potentially be a lot of us to explain and ultimately create more stress and confusion for users.  

That said, it is probably best for me to start by testing things out. @geealbers, I may follow-up separately with some questions around testing.

hbalenda:
@geealbers @Erin-Cecele The proposed change would attempt to run a `git-lfs` command to track the image directory when a user runs `quire new <project>`. If they don't have `git-lfs` installed, it will show the warning to the user with a link to the quire docs, but continue to install the project. Let me know if you want any changes made to the warning message, or I suppose another option would be to include instructions in the docs for the user to set up git-lfs on their own if they want to, and to not include this process in `quire new`.

Erin-Cecele:
Thank you @hbalenda. I had a chat with @geealbers and have a clearer understanding. I also tested somethings out. Here is the error message we suggest: 

"Warning: Git LFS (Large File Storage) is required to publish repositories with files over 100 MB to GitHub. See documentation for more info and install instructions: https://quire.getty.edu/documentation/github. This message will not impact initialization of new project directory."

I will work on documenting install and use of Git LFS and add to our docs.

Erin-Cecele:
@hbalenda I'd also like to suggest revised language for the Kindle Previwer warning message to keep things consistent. 

"Kindle Previewer is required to generate MOBI files. Please visit https://kdp.amazon.com/en_US/help/topic/G202131170 for more info and install instructions."

And just a quick note that git-lfs should also track everything in the /downloads directory. Thanks!

hbalenda:
@Erin-Cecele @garciagregg I've made the suggested changes to the warning messages, and set up git lfs to track `downloads/` and `img/`.

garciagregg:
@hbalenda I think you mean @geealbers

mphstudios (approved):

PR #81, opened 2021-03-24
Feature/DEV-8377: Disable scroll and touch zoom on entry viewer images

hbalenda:
Changes:
- Disable scroll and touch zoom on entry viewer images
- Refactor `createMap` to accept a Leaflet options object


mphstudios (approved):
this is nice

PR #82, opened 2021-03-25
Feature/DEV-8373: Fix spinner and update processing message

hbalenda:
Changes:
- Uses `execa` for image tiler command to prevent spinner from freezing
- Adds message indicating process may take a long time to complete, though the messaging will be further improved in a subsequent PR.


mphstudios (approved):
@hbalenda this approach to getting it all working again looks good. For future refactoring, when slicing the images pass the `error` thrown as an argument to the `catch` block and push the image name, path, error message onto an errors array that we can iterate through and display at the end of the process.

PR #86, opened 2021-03-26
Refactor/DEV-8370: Change iiif seed directory

hbalenda:
Changes:
- Create empty `iiif` directory in new projects
- Search for images to tile for IIIF in `static/img/iiif` instead of `static/img/iiif/images`

IIIF Config in figures.yaml would look like this:
```
  - id: "fig-2"
    label: "Figure 2"
    src: iiif/cover.png
    iiif: /img/iiif/processed/cover/info.json
    media_type: iiif
    caption: "Walker Evans. *Sons of the American Legion, Bethlehem, Pennsylvania*, 1935."
    credit: "The J. Paul Getty Museum, Los Angeles"
```

This allows `iiif` to also be an external source.

Processed images are still output to `static/img/iiif/processed`

### DIFF
--- a/packages/cli/lib/cli.js
+++ b/packages/cli/lib/cli.js
@@ -97,6 +97,10 @@ export default class CLI extends EventEmitter {
         // First Commit
         this.notice("Initializing git in the new project directory...");
         process.chdir(projectDir);
+
+        // Create empty IIIF image directory
+        fs.mkdirSync(path.join(localStarterDir, 'static', 'img', 'iiif'), { recursive: true });
+
         spawnSync("git", ["init"]);
         if (commandMissing("git-lfs")) {
           this.warn(`Warning: Git LFS (Large File Storage) is required to publish repositories with files over 100 MB to GitHub. See documentation for more info and install instructions: https://quire.getty.edu/documentation/github. This message will not impact initialization of new project directory.`);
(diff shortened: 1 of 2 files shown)


mphstudios:
@hbalenda if our goal is to have the tiling process read `figures.yaml` for the images, it makes sense to make the default directory for `info.json` and tiles `static/images/iiif/<image-name>/`, we might also infer the location of the `info.json` file rather than requiring the `iiif` key in `figures.yaml`

```
   - id: "fig-2"
     src: images/cover.png
     media_type: iiif     
     caption: "Walker Evans. *Sons of the American Legion, Bethlehem, Pennsylvania*, 1935."
     credit: "The J. Paul Getty Museum, Los Angeles"
     label: "Figure 2"
```

geealbers:
@mphstudios and @hbalenda, regarding having the tiling process read `figures.yaml` for the images, I'd like us to think that through more before implementing it, same with the idea of making the tiling process an automated part of the build rather than a process a user has to do more manually on the side. 

Right now, I'm thinking that making it a more manual process that involves putting images in a particular directory and running `quire process --iiif` actually may be advantageous to the user in keeping things straight. As a user I actually don't want to commit high-res image files into the repository, and I wouldn't want to just through them into a mixed directory with other images only to have to fish out later. I also think I'd rather just have the tiling run once and commit the files rather than have netlify do it anew everytime. I think.

I need to focus on getting all this running in Rococo before next Tuesday, and I think doing that will give me some more insight into this. Can we talk more about this after that?

mphstudios:
@geealbers refactoring to read images for slicing from the `figures.yaml` will be a separate pull request; we can think through the user workflow more.

mphstudios (approved):

PR #87, opened 2021-03-26
Add tif to supported file types for imageslice

hbalenda:
I successfully tested slicing a `.tiff`, and I am unsure of why they were not originally included on the list of supported extensions, I haven't found anything in the `go-iiif` documentation that would indicate they aren't supported.


mphstudios (approved):
Hooray for `tiff`

PR #90, opened 2021-04-06
Update version to 0.19.0

hbalenda:


mphstudios (dismissed):
@hbalenda we might also add a changelog entry for these changes, thoughts?

mphstudios (approved):

PR #95, opened 2021-04-14
Fix/DEV-8567: Fix git-lfs static file paths

hbalenda:


geealbers (approved):
Yes, works as expected now. Thanks!

PR #96, opened 2021-04-15
Fix/DEV-8489: Error when navigating to previously viewed Leaflet images in popup

hbalenda:
Changes:
- Tears down the previous Leaflet instance in the `change` event before initializing a new one. This removes the Leaflet error `Map container is already intialized`, which is shown when Leaflet tries to re-initialize the same map more than once. 

This work illuminated some issues with the `Deepzoom` and `Map` classes, which should be refactored to have the same signature (or be the same class) and also need to be cleaned up because there are asynchronous methods in the constructors, which could cause timing issues.


themes/default/source/js/popup.js
```diff
@@ -92,6 +98,22 @@ export default function(gallerySelector, mapArr) {
...
+  const tearDownMap = () => {
+    if (deepzoom) {
+      deepzoom.map.remove();
+      return;
+    }
+    if (map) {
+      map.remove();
+      return;
+    }
+  };
```
mphstudios, line 115: We definitely need to refactor this; if the two have completely different signatures we should look at using a factory pattern to configure and instantiate a single instance of either _component_ and then an adapter module to abstract the differences.

mphstudios (approved):

PR #98, opened 2021-04-19
Feature/DEV-8433: Process IIIF images one by one

hbalenda:
The diff is hard to read, but here are the changes:
- Move promise-wrapper from highest level in `imageslice.js` to the end of the file where the processing functions are actually being called
- Instead of calling the image tiling command in a for-loop, change `iiifSlice` to return the [child process and Promise returned](https://github.com/sindresorhus/execa#api)  from `execa`, and use `Promise.all()` to execute the Promises one by one.
- Update console logs with `Processed x/n images....` after each image finishes
- Remove extra promise-wrapper from `process()` in `lib/cli.js` that was doing nothing
- Update command called in CLI tests for IIIF image processing from `imageslice` -> `process --iiif`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ Project versions conform to [Semantic Versioning](https://semver.org/)
 * `Fixed`: for any bug fixes
 - `Removed`: for deprecated features removed in this release
 
+<a name="0.19.3"></a>
+
+## [0.19.3]
+
+### Changed
+
+- Process IIIF images sequentially in `process --iiif` command.
+
 <a name="0.19.2"></a>
 
 ## [0.19.2]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,25 +13,21 @@ Project versions conform to [Semantic Versioning](https://semver.org/)
 * `Fixed`: for any bug fixes
 - `Removed`: for deprecated features removed in this release
 
-<a name="0.19.3"></a>
+<a name="unreleased"></a>
 
-## [0.19.3]
+## [unreleased]
 
 ### Changed
 
 - Process IIIF images sequentially in `process --iiif` command.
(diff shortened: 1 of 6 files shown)


mphstudios (approved):
@hbalenda this looks good, definitely a minor version increment.

mphstudios (approved):

geealbers:
@hbalenda I just installed with `npm run cli:install` and tried processing iiif images and they're still processing simultaneously for me. Tried in both an old and a new project.

hbalenda:
@geealbers  hmmmmmm interesting, are you getting the new logging output, or does it just crash/time out like it did before? it should update the message as each one finishes, i.e. "processed 1/6 images..."

geealbers:
@hbalenda no. no new logging, just crashing. I did wonder if there was some way for me to verify that `npm run cli:install` worked as expected and I actually was using the newest version of the CLI when I tested. But I wasn't sure where/what to check.

hbalenda:
When it's published to npm, you can always check the version, `quire -V` but since this is an unreleased feature that won't work (@mphstudios maybe we start adding alpha versions to help with debugging in cases like this?) It does sound like you don't have the latest version running though if the logging hasn't changed (it should say "processing X images...." instead of "processing images..."), I would double check you pulled the latest changes from main before installing, and if so maybe uninstall and then reinstall, though you should only have to do that if you're switching between the npm package and the local package.

geealbers:
@hbalenda oh, it does actually say "Processing 8 images...", but it never gets to a "Processed 1/8 images..." message, and in Activity Monitor, I can see it running 8 instances of iiif-tile-seed-macos.

hbalenda:
@geealbers ah yeah I see now.. it doesn't seem like the change did what I expected it to, I'll reopen the issue.

PR #101, opened 2021-04-21
Fix/DEV-8433: Fix sequential slicing of images

hbalenda:
Replaces `Promise.all` with an anonymous async function with a for loop that `awaits` each image slice command's completion before proceeding to the next. Previously, I was using an await inside a `.map` which _does not wait for the function to resolve before moving to the next item_... all it does is allow you to execute a callback _after_ that item has completed.

### DIFF
--- a/packages/cli/lib/imageslice.js
+++ b/packages/cli/lib/imageslice.js
@@ -161,10 +161,8 @@ export default async function () {
         await iiifSlice(image).then(() => {
           imagesToSlice--;
           imagesSliced++;
-          if (imagesToSlice === 0) imagesDone();
         }).catch(() => {
           imagesToSlice--;
-          if (imagesToSlice === 0) imagesDone();
         }).finally(() => {
           spinner.start(`Processed ${imagesSliced}/${originalImages.length} ${pluralize('image', originalImages.length)}...`);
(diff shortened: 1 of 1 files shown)


geealbers:
@hbalenda this is working for me now!

mphstudios (approved):
@hbalenda I think this is good for an immediate fix however I am wary of the immediately self invoked function, 
```
async (() => {...})()
``` 

Restating a few goals may lead us to an implementation that we can reuse elsewhere:

> Images need to be processed (sliced into tiles) sequentially so that it is less likely to overload the operating system.
> Processing images sequentially simplifies the logic to display to the user the incremental progress of the processing command.

Restated, it describes something a lot like a _task queue_; each task in the queue should be a _function_ that can be invoked, and which returns a _Promise_. Using a "queue" rather than a "stack" will allow processing the queue to `start` and still add images to the queue. Other code can listen for the queue to be empty or processing to be complete to trigger other actions.

hbalenda:
Agreed. I updated [DEV-8601 (Refactor imageslice)](https://jira.getty.edu/browse/DEV-8601) with a note to replace the self-invoked function with a task queue.

PR #104, opened 2021-04-27
Refactor/DEV-7298: CircleCI OS Matrix 

hbalenda:
Changes:
- Moves `.circleci/config.yml` to the project root
- Uses an os matrix to install, build, and test quire on linux, macos, and windows.

The layout for the config.yml is a little repetitive, and I'm open to feedback, but there are a lot of annoying constraints using the matrix parameters, especially because the windows executor name has a `/` in it. I also chose not to use workspaces because of the cost of persisting and attaching workspaces was slowing down the process pretty significantly since it's such a large project.  Instead, I split the job into smaller pieces using commands.

Notes:
- Kindlepreviewer commands (expectedly) fail on linux. Need to skip epub test in that environment.
- Several tests are failing on windows. The `quire --help` output is different than expected. The epub test also exits with a million `PandocResourceNotFound`  errors. This will require some looking into.
- Weirdly, the github/circleCI integration shows the results for the windows build under `ci/circleci: default`


packages/cli/test/cli.test.js
```diff
@@ -93,22 +93,18 @@ describe("CLI", () => {
...
-      await quire.mobi(testFilePath);
-      assert.equal(
-        fs.existsSync(`${testFilePath}.mobi`),
-        true
-      );
-      done();
-    },
-    timeout
-  );
+  if (process.platform !== "linux") {
```
mphstudios, line 96: A comment here about why we are skipping this test would be kind to our future selves.

.circleci/config.yml
```diff
@@ -0,0 +1,157 @@
...
+    macos:
+      xcode: 11.4
+  win: 
+    machine:
+      image: windows-server-2019-vs2019:stable
+      resource_class: windows.medium
+      shell: powershell.exe
+
+anchors:
+  - &npm_cache_key package-cache-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
mphstudios, line 24: I recently added a `{{ .Environment.CACHE_VERSION }}` to Getty.edu so that we have a mechanism to manually invalid the package cache in CircleCI.

mphstudios (approved):
@hbalenda this is fantastic, it is going to be a relief to have the automated test suite running.

PR #113, opened 2021-04-29
DEV-8590: Update hugo-bin to 0.70.2

hbalenda:
I've reviewed the changelog, and have only made a few changes to bring the project up to date. They are:

- Update names of taxonomy "kinds" in `disableKinds`
- Simplify the method to get the epub cover image, which was adding and then regexing out `localhost:1313` which wasn't working on windows because of the path delimiters. I don't think removing this should cause any issues, it's only used to generate the epub cover. Note, the windows job still fails, but previously it was failing on the `build epub` test, and now it does on the `mobi` test. Progress! 
- Normalize relative image urls before fetching the local path to image for the epub build. See inline note.


packages/cli/lib/chapter.js
```diff
@@ -234,13 +234,14 @@ class Chapter {
...
+        const imgPath = src.replace(/^(\.+\/)+img\//, "");
         let localPath = path
           .resolve(
             path.join(chapter.config.outputDir, chapter.config.imageDir),
-            src
+            imgPath
           )
           .replace(/\/localhost:1313/, chapter.config.outputDir);
-        return "./" + localPath;
+        return localPath;
```
hbalenda, line 244: the image `src` that's from the template here uses the `relURL` filter, which adds the appropriate relative prefix to the image path - but this is relative to the directory structure of `site`, so here when we attempt to get the local path to the image, the relative path is incorrect for images in template nested more than one level deep. This change adds a regex to remove the relative part of the path (../.., ../) up to `img`, and then uses it to resolve the local path to the image.

geealbers (approved):
I tested this out using content and layout customization files from four recent projects and wasn't seeing any errors. @hbalenda is there still anything you're seeing on your end? Or are there particular things you'd like me to look at? If not, I think this is probably good to merge.

mphstudios (approved):

PR #118, opened 2021-05-10
Update License

Erin-Cecele:
Update license agreement to _Creative Commons Attribution-NonCommercial 4.0 International License_.


geealbers (commented):
I don't think we actually want this here. By publishing the license here I think it implies that it's in effect for anyone that uses the repo, rather than them having to use the sign-up form and checking the box that they agree to the license.

And because this isn't the documentation site, I don't think we need the CC license info for that either.

But maybe you had something in mind I'm not thinking of?

Erin-Cecele:
@geealbers Nope not at all. I was just trying to get everything in sync and crossed some wires with this one. I will close this one out.

Issue #136, opened 2021-06-23
Using Prince in PowerShell - cannot parse C:\Program Files (x86)


geealbers:
### Discussed in https://github.com/thegetty/quire/discussions/135

<div type='discussions-op-text'>

<sup>Originally posted by **sophielamb** June 16, 2021</sup>
I am not able to generate the static pdf files for our project. I installed Prince the same time as installing Quire on my Windows machine.

When I run `quire pdf` Prince kicks into gear and sets about generating the file, but it cannot read the program file location. On installation it went into the default `C:\\Program Files (x86)`. The spaces mean it can't be read. 

```
PS C:\Users\sophi> Set-Location -Path C:\\Users\sophi\Documents\Materia\Issue-1-Spring-21-q-01
PS C:\Users\sophi\Documents\Materia\Issue-1-Spring-21-q-01> quire pdf
\ Building PDFBrowserslist: caniuse-lite is outdated. Please run next command `npm update`
- Building PDF'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
× Error: Command failed: C:\Program Files (x86)\Prince\engine\bin\prince.exe http://localhost:1313// http://localhost:1313//contents http://localhost:1313//intro http://localhost:1313//articles/article-jd-rh-et http://localhost:1313//articles/article-gp-mla-vm http://localhost:1313//articles/article-jll-sac-fc http://localhost:1313//articles/article-my http://localhost:1313//articles/article-ej http://localhost:1313//about/about http://localhost:1313//about/acknowledgements -o static\downloads\output.pdf --no-warn-css-unsupported --no-warn-css-unknown --no-warn-css
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

PS C:\Users\sophi\Documents\Materia\Issue-1-Spring-21-q-01>
```

Is there a way to get PowerShell to recognise the location of the Prince folder? Have other win users had similar problems, or use Prince differently?

The closest workaround I could think of is to enclose the path name in quotation marks, but that's not possible in this situation.

best wishes,
Sophie

</div>

PR #137, opened 2021-06-23
fix/issue-136: Properly wrap Prince pdf command in quotes for Windows users

geealbers:
For Windows users using PowerShell, which is now the default shell application on Windows, the Prince command needs to have the part of its path that includes spaces wrapped in double quotes. The following is the format that works (in PowerShell as well as in the older Command Prompt application), and is what's being done in this PR:

✅

```
$ C:\"Program Files (x86)"\Prince\engine\bin\prince.exe --version
```

FYI, these variations, wrapping the whole path in double quotes or wrapping the one part of the path in single quotes, have been tested and **do not work**:

🚫

```
$ "C:\Program Files (x86)\Prince\engine\bin\prince.exe" --version
```

```
$ C:\'Program Files (x86)'\Prince\engine\bin\prince.exe --version
```

---

This PR fixes issue #136, which was brought to our attention by Quire user @sophielamb. Thanks Sophie!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Project versions conform to [Semantic Versioning](https://semver.org/)
 
 ### Fixed
 - Include label in image modal caption/credit block
+- Wrap Windows Prince command in quotes
 
 <a name="0.19.1"></a>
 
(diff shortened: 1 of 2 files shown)


hbalenda (approved):
@geealbers this looks good to me - I'm going to fix the failing tests in a separate PR (unrelated)

geealbers:
Thanks @hbalenda, merging in

PR #155, opened 2021-08-11
Docs/update suggestion box template

Erin-Cecele:
I think this works a lot better. Let me know what you think!


geealbers (approved):
Great!

PR #162, opened 2021-08-23
Feature/DEV-8604: Implementation of multi-level subsections

geealbers:

### Checklist

- [X] I have read the [CONTRIBUTING.md](https://github.com/thegetty/quire/blob/main/CONTRIBUTING.md) file.

- [ ] I am requesting feedback on a draft pull request

[DEV-8604](https://jira.getty.edu/browse/DEV-8604): "Implementation of multi-level subsections in Quire"

The goal is to support publications with multiple levels of nested sub-sections, and more specifically to generate and display a valid contents list showing and linking to these nested pages in both contents pages and the sidebar menu, and in all contents list formats Quire supports (list, brief, abstract, and grid). The list also needs to support section pages with and without landing pages of their own (ie., linked and not linked).

The primary work was in rewriting the contents-list.html partial, as well as the quire-contents-list.scss styles. For the partial, I split it into multiple smaller partials to aid in clarity and ease future development. They're all in the `layouts/partials/contents-list directory`. The primary partial is `contents-list.html`, there are also a set of four `list-item-types` partials (abstract, brief, grid, and list), and finally four “utility” partials (add-card-image, add-closing-tags, add-list-item-type, add-section-heading). I'd be interested in the team's thoughts on the naming of these utility partials as we haven’t done something like this before.

For the SCSS, I rewrote it and ultimately cut the lines down by about 50% and hopefully made them a little more legible. For the grid layout, I relied more on CSS Grid, where before we used floats and grid only where supported. Support seems wide enough now. I tested the styles in multiple browsers and in the PDF output using two different sets of test content with various configurations and in both the modern and classic theme styles.

Finally, I removed two parameters from the default config.yml file: `tocType` and `menuType`. I then added a new parameter `menuClass`. The original params were meant to offer some customization options but tocType was redundant to and sometimes in conflict with the classes we instructed users to add to the page yaml for contents pages so it’s not needed at all. And replacing menuType (whose possible values were "short" and "full") with menuClass ("list" and "brief") offered the chance to make the menu customization options more parallel with that of the contents pages.

Yes, the parameter changes mentioned above will have to be documented. I also think we could do a better job documenting contents functionality in general. I’ll work on this.

Here are some screenshots of test content in multi-level sub-sections in all four contents style types (brief, list, abstract, and grid):

content-list-demo


geealbers:
@mphstudios, @hbalenda @anderspollack This is a significant PR for this new feature. I've added all three of you as reviewers as we've done in the past, but trust you'll work out how best to approach this. Happy to chat it through on our call tomorrow as well.

This also our first use of the Pull Request Template that @Erin-Cecele created for us. Feedback on that would also be most welcome: https://github.com/thegetty/quire/blob/main/.github/pull_request_template.md. Does it give you the right info? Too much / too little? Anything else you'd like contributors to be including?

Thanks!

mphstudios (approved):
@geealbers removing the `add-` prefix from the `content-list` partials may make more sense. Otherwise this looks good overall.

geealbers:
Thanks @mphstudios! I went ahead and removed the `add-` prefix as you suggested. Looking at it again with some distance I think you're right. With the new commit, you'll need to approve the PR again before I can merge

mphstudios (approved):

PR #214, opened 2021-11-18
Set markdownTemplateEngine to nunjucks

mphstudios:
Sets the default engine used to pre-process markdown templates to nunjucks in order to support keyword args in shortcodes, for example:

```
{% q-figure id="fig-1", class="pulled-right" %}
```

Nota bene: this has implications on the shortcode method signatures. Some Quire shortcodes use positional arguments, for example `q-cite`, and others use keyword arguments. Refactoring the shortcode method signatures might be refactored to support both positional and keyword arguments. Alternately, all shortcodes might be refactored to require keyword arguments.


hbalenda (approved):

hbalenda:
🎉

PR #246, opened 2022-01-27
Refactor shortcode components signatures

hbalenda:
Changes:
- Change signature in `plugins/components/index.js`
- Refactor all components to use the same signature

There is still a problem here - if you log `this.page` when each component is registered, it only has value on the components that are used inside a markdown template or a layout. Components used inside components don't have access to the page context.


hbalenda:
Currently not working since `this.page` is not available when registering the components, though it is when we register the shortcodes. Troubleshooting why `this` might be different in those two cases.

anderspollack (approved):

PR #286, opened 2022-04-20
Quire v1.0.0-pre-release

mphstudios:
Feature branch for the Quire v1.0 pre-release. Quire v1.0 makes significant changes to the Quire project, notably replacing binary dependencies for generating publication outputs with NodeJS libraries. Hugo is replaced with Eleventy for generating publication HTML output, while Pandoc and PrinceXML are replaced with NodeJS libraries for generating EPUB and PDF output.


Erin-Cecele (dismissed):

geealbers (dismissed):
Here. We. Go! 🚀

PR #287, opened 2022-04-21
Feature/canvas panel shortcode

hbalenda:
Changes:
- Include canvas-panel web component
- Create `canvasPanel` shortcode that renders a `figure` from `figures.yaml` 
- Adds async support to `addComponentTag`


packages/11ty/_plugins/components/addComponentTag.js
```diff
@@ -19,9 +19,15 @@ module.exports = function(eleventyConfig, component, tagName) {
...
-  })
+  if (component.constructor.name === 'AsyncFunction') {
+    eleventyConfig.addJavaScriptFunction(tagName, async function(...args) {
+      return await component(eleventyConfig, globalData, { page: this.page })(...args)
+    })
+  } else {
+    eleventyConfig.addJavaScriptFunction(tagName, function(...args) {
+      return component(eleventyConfig, globalData, { page: this.page })(...args)
+    })
+  }
```
mphstudios, line 30:
```suggestion
  eleventyConfig.addJavaScriptFunction(tagName, async function(...args) {
    return (component.constructor.name === 'AsyncFunction')
      ? await component(eleventyConfig, globalData, { page: this.page })(...args)
      : component(eleventyConfig, globalData, { page: this.page })(...args)
  })
```

packages/11ty/_plugins/shortcodes/canvasPanel.js
```diff
@@ -0,0 +1,65 @@
...
+      for (const body of bodies) {
+        const { items, type } = body
+        if (type === 'Choice') {
+          return items.map(({ id }) => vault.get(id))
+        }
+      }
+    }
+  }).filter(item => item)
+}
+module.exports = function(eleventyConfig, globalData) {
```
mphstudios, line 27:
```suggestion
/**
 * Very simplified method to get choices - expects a valid manifest where choices all have identifiers
 * @todo replace with vault helper
 */
const getChoices = (annotations=[]) => {
  return annotations.flatMap(({ id, type }) => {
    const annotation = vault.get(id)
    if (annotation.motivation.includes('painting')) {
      const bodies = vault.get(annotation.body)
      for (const body of bodies) {
        const { items, type } = body
        if (type === 'Choice') {
          return items.map(({ id }) => vault.get(id))
        }
      }
    }
  }).filter(item => item)
}

/**
 * A shortcode that renders the canvas-panel web component
 */
module.exports = function(eleventyConfig, globalData) {
```

packages/11ty/_plugins/components/addComponentTag.js
hbalenda: This will make every component an async function

packages/11ty/_includes/components/figure/image.js
```diff
@@ -10,21 +10,34 @@ const path = require('path')
...
+    const hasManifestAndCanvasIds = (!!canvasId && !!manifestId) || !!iiifContent || !!choices
+
+    let imageElement;
+
+    switch (true) {
+      case hasManifestAndCanvasIds:
+        imageElement = await canvasPanel({ canvasId, id, manifestId, preset })
+        break;
+      case hasTiles:
+        imageElement = `<image-service alt="${alt}" class="q-figure__image" src="${imageSrc}"></image-service>`
```
mphstudios, line 35:
```suggestion
        imageElement = `<image-service alt="${alt}" class="q-figure__image" src="${imageSrc}" />`
```

packages/11ty/_includes/components/figure/image.js
```diff
@@ -10,21 +10,34 @@ const path = require('path')
...
-      <img alt="${alt}" class="q-figure__image" src="${imageSrc}" />
-    `
+    const srcParts = src.split(path.sep)
+    const hasTiles = srcParts[srcParts.length - 1] === 'tiles'
+    const hasManifestAndCanvasIds = (!!canvasId && !!manifestId) || !!iiifContent || !!choices
+
+    let imageElement;
+
+    switch (true) {
+      case hasManifestAndCanvasIds:
```
mphstudios, line 31: Since this is also `true` when there is `iiifConent` or `choices` perhaps this should be named `hasCanvasPanelProps` or `useCanvasPanel`
```suggestion
    const hasCanvasPanelProps = (!!canvasId && !!manifestId) || !!iiifContent || !!choices

    let imageElement;

    switch (true) {
      case hasCanvasPanelProps:
```

packages/11ty/_plugins/components/addComponentTag.js
```diff
@@ -19,9 +19,13 @@ module.exports = function(eleventyConfig, component, tagName) {
...
-  eleventyConfig.addJavaScriptFunction(tagName, function(...args) {
-    return component(eleventyConfig, globalData, { page: this.page })(...args)
-  })
+  component.constructor.name === "AsyncFunction"
+    ? eleventyConfig.addJavaScriptFunction(tagName, async function (...args) {
+        return await component(eleventyConfig, globalData, { page: this.page })(...args)
+      })
+    : eleventyConfig.addJavaScriptFunction(tagName, function (...args) {
+        return component(eleventyConfig, globalData, { page: this.page })(...args)
+      })
```
mphstudios, line 28: Using an _expression_ without a left-hand assignment might not be parsed well, if moving the ternary to the `return` statement does not work it may read better and be parser safer as you had written it using an `if/else` statement.

packages/11ty/_plugins/shortcodes/canvasPanel.js
```diff
@@ -0,0 +1,101 @@
+/**
+ * CanvasPanel shortcode that renders the digirati <canvas-panel> web component
+ * @see {@link https://iiif-canvas-panel.netlify.app/docs/intro/ Canvas Panel Documentation}
+ * @todo import iiif config and use to set canvasId and manifestIds for figures.yaml-generated manifests
+ */
+const { html } = require('common-tags')
+const { globalVault } = require('@iiif/vault')
```
mphstudios, line 7:
```suggestion
const { globalVault } = require('@iiif/vault')
const { html } = require('common-tags')
```

packages/11ty/_plugins/shortcodes/canvasPanel.js
```diff
@@ -0,0 +1,101 @@
...
+   */
+  return async function(params) {
+    let { canvasId, choiceId, id, manifestId, preset } = params
+
+    switch(true) {
+      case !!manifestId && !!canvasId:
+        break;
+      case !!id:
+        canvasId = [process.env.URL, '_assets', 'images', '_iiif', id, 'canvas'].join('/')
+        manifestId = [process.env.URL, '_assets', 'images', '_iiif', id, 'manifest.json'].join('/')
```
mphstudios, line 50: Is using `path.join()` here more idiomatic?

packages/11ty/_plugins/shortcodes/canvasPanel.js
```diff
@@ -0,0 +1,101 @@
...
+
+    const hasChoices = !!choices.length
+
+    choicesUI = choices.map((item, index) => {
+      const classes = ['canvas-choice']
+      if (index === 0) classes.push('canvas-choice--active')
+      return `
+        <button class="${classes.join(' ')}" type="button" value="${item.id}">
+          ${item.label.en ? item.label.en[0] : item.label}
+        </button>
```
mphstudios, line 85: Implementing the choice interface as a component would allow using either `<button>` elements or a group of `<input />` and `<label>` elements, for example
```
<input class="${classes.join(' ')}" type="radio" id="item.id" name="choice" value="${item.id}" checked>
<label for="${item.id}">${item.label.en ? item.label.en[0] : item.label}</label>
```

PR #296, opened 2022-04-28
Refactor: Global Data

hbalenda:
Changes:
- Add data in all yaml files in `_assets/_data` to `eleventyConfig.globalData`
- Fixes issue where previously you would need to restart the server to see changes to yaml data reflected in the project

Maybe I'm forgetting a very good reason we didn't do this to begin with when we refactored the shortcodes, but I was trying to figure out how to access data from one of the yaml files from the IIIF plugin, and adding them to `eleventyConfig.globalData` using [`addGlobalData`](https://www.11ty.dev/docs/data-global-custom/) seemed like the simplest way to make that data available to all plugins. This change does make destructuring global data in shortcodes very verbose.


mphstudios (approved):
this make more sense, let's merge it!

PR #304, opened 2022-05-09
Refactor/navigation dev 12067

hbalenda:
Quire Table of Contents with `eleventyNavigation`

Note: This only addresses the TOC data, styles will be addressed in another PR.

Changes:
- You no longer have to specify a `section` in the frontmatter when a page is nested under that section's directory.
- If you must include a page that is outside a section directory in a section's TOC, you can specify `parent: <section-name>` in the front matter. Section name will be the directory path relative to the content directory without beginning and ending slashes, i.e. "/catalogue/foo" would be "catalogue/foo"
- All sections must have an `index.md`. If the section does not have a landing page, the `index.md` should include the title of the section and `online: false` in the front matter.
- Rename `contents` layout `table-of-contents`
- Rename table-of-contents layout `class` frontmatter key to `presentation` 

- Adds `tableOfContents` collection, which includes all pages where `toc` does not equal `false`
- Adds `menu` collection, which includes all pages where `menu` does not equal `false`
- Refactors menu and TOC to use collections and `eleventyNavigation`


packages/11ty/content/_data/eleventyComputed.js
```diff
@@ -25,20 +25,39 @@ module.exports = {
...
+        layout: data.layout,
+        object: data.object,
+        online: data.online,
+        short_title: data.short_title,
+        subtitle: data.subtitle,
+        summary: data.summary,
+        title: data.title,
+        weight: data.weight
+      }
+    },
```
hbalenda, line 48: The alternative to this is providing the `tableOfContents/list` with `pages` and the navigation object to render and then looking up the full `page` object as we render each TOC item. `eleventyNavigation` items only include select page data (title and url), not everything we need to render quire TOCs which can include contributors, abstracts, etc.

packages/11ty/_includes/components/table-of-contents/list.js
```diff
@@ -0,0 +1,40 @@
...
+  const eleventyNavigation = eleventyConfig.getFilter('eleventyNavigation')
+  const tableOfContentsItem = eleventyConfig.getFilter('tableOfContentsItem')
+
+  return function(params) {
+    const { navigation, type } = params
+
+    const renderList = (pages) => {
+      return html`
+        <ul>
+          ${pages.map(page => {
```
mphstudios, line 24: linting arrow function args
```suggestion
          ${pages.map((page) => {
```

packages/11ty/_layouts/table-of-contents.11ty.js
```diff
@@ -35,63 +33,22 @@ module.exports = class TableOfContents {
...
-        }
-      })
+    /**
+     * The pages to include in the table of contents
+     * Either the project or a project section
+     */
+    const currentNavigationItem = this.eleventyNavigation(collections.tableOfContents).find(
+      ({ url }) => url === page.url
+    )
+    const isSection = !!currentNavigationItem.children && currentNavigationItem.children.length > 0
```
mphstudios, line 48: Can this be written as below?
```suggestion
    const isSection = currentNavigationItem.children && currentNavigationItem.children.length
```

packages/11ty/_layouts/table-of-contents.11ty.js
```diff
@@ -35,63 +33,22 @@ module.exports = class TableOfContents {
...
+    const navigation = isSection 
+      ? this.eleventyNavigation(collections.tableOfContents).find(({ url }) => url === page.url).children
```
mphstudios, line 50: We might make a helper function from the following expression so that we need not repeat it
```
this.eleventyNavigation(collections.tableOfContents).find(({ url }) => url === page.url)
```

packages/11ty/content/_data/eleventyComputed.js
mphstudios: @hbalenda let's use this implementation, it is nicely explicit.

packages/11ty/_includes/components/menu/list.js
```diff
@@ -10,35 +10,24 @@ module.exports = function(eleventyConfig) {
...
+          listItem += `<li class="section-item">${menuItem(item)}`
           if (config.params.tocType === 'full') {
-            subListItems = pages
-              .filter((item) => item.data.section === page.data.section && item.data.layout !== 'table-of-contents')
-              .map((item) => {
-                if (page.fileSlug !== item.fileSlug)
-                  return `<li class="page-item">${menuItem({ page: item.data })}</li>`
-              })
-            listItem += `<ul>${subListItems.join('')}</ul>`
+            listItem+= `<ul>${renderList(item.children)}</ul>`
```
mphstudios, line 23:
```suggestion
            listItem += `<ul>${renderList(item.children)}</ul>`
```

packages/11ty/_layouts/base.11ty.js
```diff
@@ -27,7 +27,7 @@ module.exports = function(data) {
             aria-expanded="false"
             role="contentinfo"
           >
-            ${this.menu({ imageDir, pageData, pages, publication})}
+            ${this.menu({ pageData, collections, publication})}
```
mphstudios, line 30: for alphabetic consistency
```suggestion
            ${this.menu({ collections, pageData, publication})}
```

packages/11ty/_layouts/table-of-contents.11ty.js
hbalenda: not if we want it to evaluate to true or false, your suggestion will evaluate to the the length of the array

packages/11ty/content/_data/eleventyComputed.js
```diff
@@ -25,20 +25,39 @@ module.exports = {
...
+    data: (data) => {
+      return {
+        abstract: data.abstract,
+        contributor: data.contributor,
+        figure: data.figure,
+        image: data.image,
+        label: data.label,
+        layout: data.layout,
+        object: data.object,
+        online: data.online,
```
mphstudios, line 41: Had we discussed renaming `online` to something else?

packages/11ty/_layouts/table-of-contents.11ty.js
mphstudios: For whatever reason checking for length greater than 0 seems to set a boolean has always struct me as odd, if you think either of below are less so, otherwise it may be my thinking that is odd
```
!!currentNavigationItem.children && !!currentNavigationItem.children.length
```
or
```
!!(currentNavigationItem.children && currentNavigationItem.children.length)
```

packages/11ty/content/contents.md
```diff
@@ -1,7 +1,7 @@
 ---
-class: list
 layout: table-of-contents
 search: false
 title: Contents
+type: list
```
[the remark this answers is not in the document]
hbalenda: hmmm true, maybe presentation?

packages/11ty/content/_data/eleventyComputed.js
hbalenda: We did but we didn't come up with anything, I'm find with leaving it for now

packages/11ty/content/contents.md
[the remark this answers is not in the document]
geealbers: Just to chime in, from a user's perspective, I think having a consistent name for "when I want to change the type of this `layout` to some particular variant" would be better than highly specific names each time for each different layout. `presentation` might be good. `variant`, `layout_type`?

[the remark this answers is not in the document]
mphstudios: `presentation` is nice.

PR #306, opened 2022-05-10
Refactor: IIIF Data Fetching

hbalenda:
I think this will help with the lightbox 💡

Changes:
- Move all vault fetching to `_plugins/iiif/addGlobalData`, which iterates over figures.yaml and adds IIIF `manifest`, `canvas`, `choiceId` and `choices` to `globalData.figures`

### DIFF
--- a/packages/11ty/_includes/components/figure/choices.js
+++ b/packages/11ty/_includes/components/figure/choices.js
@@ -1,39 +1,7 @@
-const { globalVault } = require('@iiif/vault')
-const vault = globalVault()
-
-/**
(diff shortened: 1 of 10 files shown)


mphstudios (commented):
@hbalenda @anderspollack what do you think about encapsulating the current logic from `isImageService`, `hasCanvasPanelProps`, and the `figure/image` `switch` statement, into a module that exports a `getFigureImageElement(figure, options)` method?

mphstudios (approved):

PR #312, opened 2022-05-11
Fix: Table of contents nesting & presentation

hbalenda:
Fixes: 
- Data for brief, abstract, and grid presentations
- Deeply nested sections in TOC

Old quire publication with proper styles and functionality: https://shiny-belekoy-51611c.netlify.app/contents-abstract/

### DIFF
--- a/packages/11ty/_includes/components/table-of-contents/grid-item.js
+++ b/packages/11ty/_includes/components/table-of-contents/grid-item.js
@@ -83,7 +83,7 @@ module.exports = function (eleventyConfig) {
         break;
     }
 
-    if (!children || (isOnline && classes.includes('level-0') && !!children)) {
+    if (!children) {
       mainElement = `
         <div class="card ${imageAttribute} ${slugPageAttribute}">
           ${imageElement}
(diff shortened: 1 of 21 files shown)


anderspollack:
Ok I think these styles are reasonably fixed. The nested typography matches pretty well I think. Some smaller breakpoints of `grid` contents don't have the exact same margins, since they seem to rely on some more global `.quire-page__content` styles and I don't wanna inadvertently break some other pages. But nothing looks broken

packages/11ty/_includes/components/table-of-contents/grid-item.js
```diff
@@ -0,0 +1,119 @@
...
+    const slugPageAttribute = children ? "slug-page" : ""
+
+    let imageElement
+    switch (true) {
+      case !!image:
+        imageElement = `<div class="card-image">
+            <figure class="image">
+              <img src="${path.join(imageDir, image)}" alt="" />
+            </figure>
+          </div>`
```
mphstudios, line 69: Consistently using a `common-tags` method may produce better formatted markup?
```suggestion
        imageElement = html`
          <div class="card-image">
            <figure class="image">
              <img src="${path.join(imageDir, image)}" alt="" />
            </figure>
          </div>
        `
```

mphstudios (approved):

PR #313, opened 2022-05-11
Feature: Contributor variants

hbalenda:
To do:
- Fix citation/contributor component for >= 3 contributors: DEV-12168

Changes:
- Refactors `contributor` shortcode and `contributorList` component as `contributors` shortcode with formats: `string`, `name`, `name-title`, `name-title-block`, `bio`
- Adds additional variant `name-title-block` to `contributors` shortcode
- Refactors `getContributor` to accept contributor object
- Adds `eleventyComputed` properties `pageContributors` and `publicationContributors`, so usage of the shortcode changes for example from `{% contributors type="page" %}` to `{% contributors contributors=pageContributors%}`
- Uses `contributor_as_it_appears` as override for page or publication contributors in eleventyComputed
- Sets default contributors format from `contributor_byline` defined on page or in `config.yaml`


packages/11ty/content/_data/eleventyComputed.js
```diff
@@ -119,5 +109,36 @@ module.exports = {
...
+          ({ data }) => {
+            if (!data.contributor) return
+            return Array.isArray(data.contributor)
+              ? data.contributor.find(
+                (pageContributor) => pageContributor.id === item.id
+              )
+              : data.contributor.id === item.id
+          }
+        )
+        return item
```
mphstudios, line 141: @hbalenda refactoring to rewrite the nested `filter` `map` `filter` and ternaries as separate functions may improved readability though if we could understand and fix the duplicate contributors that too would help.

hbalenda: yeah, worth figuring this out before merging. i will investigate

mphstudios (approved):

mphstudios (approved):

PR #317, opened 2022-05-19
Refactor: Global data

hbalenda:
Fixes: 
- Duplicate array values in merged properties in data cascade. This is current default Eleventy behavior, and they are to discuss adding an option to dedupe: https://github.com/11ty/eleventy/issues/2393

Changes:
- Change eleventy data directory to `_computed-data` and move `eleventyComputed.js` into `_computed-data`. This means that files in `_data` will now only be loaded using the configuration API
- Update `globalData` plugin to load json files
- Update README


anderspollack (approved):

packages/11ty/content/_data/README.md
```diff
@@ -1,22 +1,10 @@
...
-  publication: {},
-  references: {
-    entries: []
-  }
-}
-```
+Data files included in the `_data` directory will be added to the project's global data using [Eleventy's configuration API](https://www.11ty.dev/docs/data-global-custom/), which allows the data to be universally accessed from Quire shortcodes and components.
+
+### Computed Data
+Complex data properties that are computed from other global data values are set in `eleventyComputed`. These values are only directly available in layouts and templates. Computed properties must be passed explicitly to shortcodes and components.
```
mphstudios, line 8:
```suggestion
Data properties that are computed from other global data values are set in `eleventyComputed`. These values are directly available only in layouts and templates; **computed data properties must be explicitly passed to shortcodes and components.**
```

PR #324, opened 2022-05-24
Fix/duplicate weights

geealbers:
@thegetty/int-dev when I was doing some testing, I noticed that there were quire a few duplicated `weights` in the Markdown pages of the starter content, which made id difficult to test the contents list stuff in particular. I fixed them here.

There was also a test in Hugo that would look for duplicate weights and post a message to users in the Terminal warning them of this and listing the pages in question. It's otherwise an insidious error for the average user to track down. Is this kind of test something we could add into the 11ty version as well.

### DIFF
--- a/packages/11ty/content/about.md
+++ b/packages/11ty/content/about.md
@@ -1,12 +1,12 @@
 ---
 title: About
-weight: 203
+weight: 503
 layout: page
 ---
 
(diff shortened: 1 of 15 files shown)


geealbers:
Ooops, selected the wrong branch to merge into. Just fixed it.

mphstudios (approved):

PR #341, opened 2022-06-07
Refactor: Collections and output types

hbalenda:
Fixes:
- Build errors on pages to be built for `pdf` and/or `epub` but not `html`

Changes:
- Refactor frontmatter to expect `outputs` key, which can be a String or Array with possible values `epub|pdf|html|none|false`
- Set output type in .env `QUIRE_OUTPUT`
- Adds a `current` collection that only includes pages that should be included in the current output type being built, specified in `.env.QUIRE_OUTPUT`
- Add computed `permalink` property to `eleventyComputed`

While excluding a page from a collection is straightforward, excluding a page from the eleventy build is less so. The consensus seems to be that setting `permalink: false` is the best way to do this, and I've added the computed permalink to `eleventyComputed` to do so, but I also had to add various data checks to prevent build failures when it seems like eleventy begins to attempt to build excluded pages, even though they are... excluded. However, the excluded pages do not exist in the final output.


packages/11ty/.env.example
```diff
@@ -1,3 +1,4 @@
 ELEVENTY_ENV=development
 ELEVENTY_OUTPUT_DIR=_site
+OUTPUT=html
```
mphstudios, line 3: Does it make this variable more explicit to add a `QUIRE_` prefix?
```suggestion
QUIRE_OUTPUT=html
```

packages/11ty/_includes/components/navigation.js
```diff
@@ -20,6 +20,7 @@ module.exports = function(eleventyConfig) {
   return function (params) {
     const { pages, pagination, title } = params
     const { currentPage, currentPageIndex, nextPage, previousPage } = pagination
+    if (!currentPage) return
```
mphstudios, line 23:
```suggestion

    if (!currentPage) return
    
```

packages/11ty/_plugins/collections/index.js
```diff
@@ -1,23 +1,35 @@
...
  */
 module.exports = function (eleventyConfig, options = {}) {
+  /**
+   * Collection of pages for the current output (epub, html, or pdf)
+   */
+  eleventyConfig.addCollection('current', function (collectionApi) {
+    return collectionApi.getAll().filter(function (page) {
+      const { outputs } = page.data
+      return shouldBuildPage(outputs)
+    });
```
mphstudios, line 15: the function passed to `filter` could be an anonymous function.

mphstudios (changes requested):
@hbalenda this is a nicely implemented solution. Were we to refactor `helpers/should-build-page` as a `helpers/filter-pages` or `helpers/page-filters` module that included methods to filter for toc pages and menu pages it would collect shared concerns; your thoughts?

hbalenda:
Unfortunately these changes now exclude sections without landing pages from the menu and TOC. Still thinking about a solution for this. I'm not sure what the most logical way to specify in frontmatter that you want a template to be represented in the menu or TOC, but not built into a page, when you also want to be able to specify that a template shouldn't be built or represented in the menu. Seems like we might need a special key for section indexes without landing pages.

mphstudios:
> Unfortunately these changes now exclude sections without landing pages from the menu and TOC. Still thinking about a solution for this. I'm not sure what the most logical way to specify in frontmatter that you want a template to be represented in the menu or TOC, but not built into a page, when you also want to be able to specify that a template shouldn't be built or represented in the menu. Seems like we might need a special key for section indexes without landing pages.

Per our discussion, something like:

```yaml
---
title: Subsection in Table of Contents
outputs: none
toc: true
---

Nota bene: for pages that are not excluded from outputs the default is `toc: true` however you may want a non-linked subsection heading to appear in the table of contents.

When a template front-matter sets `outputs` to `none` or `false` the page will be excluded from the build and not appear in the table of contents. To include a non-built page as a subsection heading in the table of contents, set `outputs: none` *and* `toc: true`

```

packages/11ty/helpers/page-filters.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Returns `true` if page should be built
+ * for the current output type defined in `env.QUIRE_OUTPUT`
+ * @param  {Object} data Eleventy `page.data` object
+ * @return {Boolean}
+ */
+const buildFilter = function ({ outputs }) {
```
mphstudios, line 7: @hbalenda `buildFilter` might be a method that _builds_ a filter (which this method does, however the other two methods are named for the things the filter function returns) should this method be named `outputsFilter`?

hbalenda: hrmm true, `currentOutputFilter` ?

packages/11ty/helpers/page-filters.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Returns `true` if page should be included in the menu
+ * @param  {Object} page Eleventy `page` object
+ * @return {Boolean}
+ */
+const menuFilter = function(data) {
+  const { menu, type } = data
+  return (currentOutputFilter(data) || menu === true) && menu !== false && type !== 'data'
+}
```
mphstudios, line 23: In `.eleventy.js`
```
eleventyConfig.addCollection('menu', pageFilters.menuFilter(collectionApi))
```
though it may be more encapsulated to init the `pageFilters` module with `collectionApi` as an arg, like
```
eleventyConfig.addCollection('menu', pageFilters(collectionApi).menuFilter())
```
In `pageFilters`, using the second approach above something like,
```
const pages = const pages = collectionApi.getAll()
const menuFilter = function() {
  const currentOutput = currentOutputFilter(pages)
  return // filtered menu items
}
```

hbalenda: Leaving for further refactoring - our use of collections and filters seems in flux still, not sure what the best implementation will be.

mphstudios (approved):

PR #342, opened 2022-06-07
Change Menu and Table of Contents list element to ordered list

hbalenda:
Some of the `ul` styles were relying on bulma, so had to bop into the styling a little bit.

### DIFF
--- a/packages/11ty/_includes/components/menu/list.js
+++ b/packages/11ty/_includes/components/menu/list.js
@@ -1,3 +1,4 @@
+const { html } = require('common-tags')
 /**
  * Renders the menu list
  *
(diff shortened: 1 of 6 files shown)


packages/11ty/_includes/components/menu/list.js
```diff
@@ -13,19 +14,23 @@ module.exports = function(eleventyConfig) {
...
-          listItem += `<li class="section-item">${menuItem(item)}`
-          if (config.params.menuType !== 'brief') {
-            listItem += `<ul>${renderList(item.children)}</ul>`
-          }
-          listItem += '</li>'
-          return listItem
-        }
-      }).join('')
+      return html`
+        <ol>
```
mphstudios, line 18: We might use the semantic `<menu>` element for this component, for reference https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu

mphstudios (approved):

PR #361, opened 2022-06-21
Refactor: IIIF Output

hbalenda:
Fixes:
Slow preview with many IIIF images by making it possible to specify a IIIF input directory other than `_assets/images/figures` for IIIF images, and outputting IIIF images to a directory outside content so that images aren't watched. For an image directory to not be watched, it needs to be outside the `content` directory.

Changes:
- Move IIIF output directory to root level
- Add `inputDir` to `iiif/config.js`
- Add template path to the error message for when a figure id is not found in `figures.yaml`
- Rename `iiif/config.js#output` -> `iiif/config.js#outputDir`
- Refactor all IIIF processing methods to expect `filename` and use `iiifConfig.inputDir` and `iiifConfig.outputDir`

Note: 
Commenting out the `entry-figure` template, which needs to be refactored to use the figure shortcode.

### DIFF
--- a/packages/11ty/_plugins/iiif/process/createImage.js
+++ b/packages/11ty/_plugins/iiif/process/createImage.js
@@ -24,7 +24,7 @@ module.exports = (eleventyConfig) => {
     const ext = path.parse(filename).ext
     const id = path.parse(filename).name
     const inputPath = path.join(inputDir, filename)
-    const outputPath = path.join(outputDir, id, filename)
+    const outputPath = path.join(outputDir, id, `${name}${ext}`)
 
     fs.ensureDirSync(path.join(outputDir, id))
 
--- a/packages/11ty/_plugins/iiif/process/createImage.js
+++ b/packages/11ty/_plugins/iiif/process/createImage.js
@@ -19,28 +19,26 @@ module.exports = (eleventyConfig) => {
    */
   return async (filename, transformation = {}, options) => {
     const { debug, lazy } = options
-    const { name, resize } = transformation
 
-    const ext = path.parse(filename).ext
-    const id = path.parse(filename).name
+    const { ext, name } = path.parse(filename)
     const inputPath = path.join(inputDir, filename)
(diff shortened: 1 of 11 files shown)


packages/11ty/_plugins/iiif/process/createImage.js
```diff
@@ -20,23 +17,22 @@ module.exports = (eleventyConfig) => {
...
+  return async (filename, transformation = {}, options) => {
     const { debug, lazy } = options
     const { name, resize } = transformation
-    const outputDir = path.join(root, output)
 
-    const ext = path.parse(input).ext
-    const id = path.parse(input).name
-    const filename = `${name}${ext}`
+    const ext = path.parse(filename).ext
+    const id = path.parse(filename).name
```
mphstudios, line 25:
```suggestion
    const { ext, name } = path.parse(filename)
```
or 
```
const { ext, name: id } = path.parse(filename)
```

packages/11ty/_plugins/shortcodes/figure.js
```diff
@@ -34,7 +34,11 @@ module.exports = function (eleventyConfig, { page }) {
     /**
      * Merge figures.yaml data and additional params
      */
-    let figure = id ? getFigure(id) : {}
+    let figure = getFigure(id)
+    if (!figure) {
+      console.warn(`Warning: The figure id "${id}" was found in the template "${page.inputPath}", but is not defined in "figures.yaml"`)
```
mphstudios, line 39: We might start making our console message fancy to help them stand out to user, perhaps something like
```suggestion
      const message = chalk.yellow(`Warning: The figure id "${id}" was found in the template "${page.inputPath}", but is not defined in "figures.yaml"`)
      console.warn(message)
```

mphstudios: Let's make the chalk drawing part of a separate feature pull-request...

mphstudios (approved):

PR #364, opened 2022-06-23
Implement logging concerns and fanciness

mphstudios:
Implementation of logging concerns for consistent console output, specifically output from the Quire shortcode components  when running `dev` and `build`


packages/11ty/_lib/chalk/index.js
```diff
@@ -0,0 +1,22 @@
+const chalk = require('chalk')
+
+module.exports = function(context='') {
```
mphstudios, line 3: We might rename this argument to `prefix` so that it is not confused with the Eleventy rendering `context`.

packages/11ty/_plugins/iiif/process/index.js
```diff
@@ -3,8 +3,11 @@ const path = require('path')
 const addGlobalData = require('./addGlobalData')
 const initCreateImage = require('./createImage')
 const initCreateManifest = require('./createManifest')
+const initLogging = require('../../../_lib/chalk')
```
mphstudios, line 6: Does it the implementation more clear to find a name for this import that makes it implicit this is factory function?

mphstudios: `chalkFactory` is explicit though were we expand the implementation it may become less so.

PR #375, opened 2022-07-06
Import `custom.css` after all application javascript to permit style overrides

anderspollack:


packages/11ty/_includes/components/head.js
```diff
@@ -71,8 +71,13 @@ module.exports = function(eleventyConfig) {
         <script type="application/ld+json">${jsonld({ canonicalURL, page })}</script>
 
         <link rel="icon" href="/_assets/images/icons/favicon.ico" />
-        <link rel="stylesheet" href="/_assets/styles/application.scss" />
-        <link rel="stylesheet" href="/_assets/styles/custom.css" />
+        <!--
+          styles are already imported in _assets/javascript/application/index.js
+          as inline minified <style type="text/css">...</style> blocks,
```
anderspollack, line 76:
```suggestion
          and rendered as inline minified <style type="text/css">...</style> blocks,
```

mphstudios (approved):
@anderspollack we should make a task, or keep DEV-12736 open, for refactoring so that the `application.js` does not import the styles but the links are added to the `<head>` by Vite.

PR #378, opened 2022-07-07
DEV-12744: Wrap `common-tags` library

anderspollack:
- Wrap custom tag functions together with all exported `common-tags` to simplify tag fn imports
- Point all instances of `require('common-tags')` to `_lib/common-tags`

**TODO**:

handle path aliases to simplify `require`s everywhere


packages/11ty/README.md
```diff
@@ -0,0 +1,73 @@
...
+  "private": ...,
+  "version": ...,
+  "description": ...,
+  "main": ...,
+  "scripts": ...,
+  "keywords": ...,
+  "author": ...,
+  "license": ...,
+  "devDependencies": ...,
+  "dependencies": ...,
```
mphstudios, line 25:
```suggestion
  // omitted for example
```

packages/11ty/README.md
```diff
@@ -0,0 +1,73 @@
...
```
mphstudios, line 59:
```suggestion
  // omitted for example
```

packages/11ty/README.md
```diff
@@ -0,0 +1,53 @@
+## Node module aliases
+
+Quire uses `module-alias` to handle `import`/`require` aliases.
```
mphstudios, line 3:
```suggestion
Quire uses the `module-alias` package for `import` and `require` aliases. 
```

packages/11ty/README.md
```diff
@@ -0,0 +1,53 @@
...
+Quire uses `module-alias` to handle `import`/`require` aliases. 
+
+`module-alias` is registered in `.eleventy.js`:
+
+``` javascript
+require('module-alias/register')
+```
+
+It is configured under the `_moduleAliases` key in `package.json`:
+
```
mphstudios, line 12:
```suggestion

`package.json`
````

File: packages/11ty/README.md [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+``` javascript
+const { renderOneLine, stripIndent } = require('~lib/common-tags')
+```
+
+### Alternative implementation: import mapping
+
+See https://github.com/ilearnio/module-alias/issues/113
+
+To get the same results with no external dependency, we can use subpath imports, configured in the `imports` key of `package.json`
+
```
mphstudios, line 36:
```suggestion
### Node import mapping

An alternate implementation with no external dependency uses [Node.js subpath imports](https://nodejs.org/api/packages.html#subpath-imports), which are configured under the `imports` key of `package.json`
```

mphstudios (changes requested):
See README suggestions...

mphstudios (approved):

PR #393, opened 2022-07-18
DEV-12646: Ensure assets get written with same input directory structure; no hashed filenames

anderspollack:
**Note:** The function passed to `assetFileNames` also bundles assets from `node_modules` into the static build (icon fonts)

Feels a bit hacky to have to manually recreate each asset input filename to preserve the directory structure, but I could not find any other way to do it. I looked into setting `rollupOptions.output.preserveModules` https://rollupjs.org/guide/en/#outputpreservemodules but it did not work for this issue; something else in the vite plugin's defaults was in conflict

### DIFF
--- a/packages/11ty/.eleventy.js
+++ b/packages/11ty/.eleventy.js
@@ -171,7 +171,7 @@ module.exports = function(eleventyConfig) {
             }
           }
         },
-        sourcemap: 'true'
+        sourcemap: true'
       },
       /**
        * Set to false to prevent Vite from clearing the terminal screen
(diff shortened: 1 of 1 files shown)


packages/11ty/.eleventy.js
```diff
@@ -147,13 +147,29 @@ module.exports = function(eleventyConfig) {
...
+              const fullFilePathSegments = name.split('/').slice(0, -1)
+              let filePath = '_assets/';
+              ['_assets', 'node_modules'].forEach((assetDir) => {
+                if (name.includes(assetDir)) {
+                  filePath +=
+                    fullFilePathSegments
+                      .slice(fullFilePathSegments.indexOf(assetDir) + 1)
+                      .join('/') +
+                    '/'
+                }
```
mphstudios, line 168: @anderspollack we might want to rewrite this as a separate method or rollup `plugin` though let's merge this solution.

mphstudios (approved):

PR #397, opened 2022-07-19
DEV-12647: Include print stylesheet in `pdf.html`

anderspollack:
This PR uses `sass` API to manually compile stylesheets for print. It concatenates the compiled contents of `application.scss`, `print.scss`, and `custom.css`, in that order. I feel like we will probably omit the `application` and `custom` styles when `print.scss` gets fleshed out, but I leave CSS architecture things to be decided when the style overhaul work gets going.

- Manually compile stylesheets for pdf.html in `transforms/pdf/write.js`
- Ensure `print.scss` imports `colors` and `variables` so it can be an entrypoint
- In PDF layout, link to single compiled stylesheet


packages/11ty/content/_assets/styles/print.scss
```diff
@@ -1,4 +1,6 @@
 @use "sass:math";
+@import "colors";
+@import "variables";
```
mphstudios, line 3: @anderspollack we may want to use `@use` rather than `@import`

anderspollack: @mphstudios for some reason I was getting this error with `@use` -- it should be able to find `variables.scss` but `$white` is still missing somehow

```
Error: Undefined variable.
   ╷
37 │ $quire-cover-text-color: $white;
   │                          ^^^^^^
   ╵
  .11ty-vite/_assets/styles/variables.scss 37:26   @use
```

PR #402, opened 2022-07-20
Refactor: page classes & main element

hbalenda:
Changes:
- Moves `<main>` element into `base` layout
- Add page classes to frontmatter `class` property, which will be applied to `<main>` in the `base` layout
- Replaces `main` with `section` tag for pdf output

Fixes:
- Defining page classes in frontmatter
- Adds `page-one` and `frontmatter` (the other kind of frontmatter) classes to correct pages
- Redundant wrapper `section` tag on pages in pdf output

Note: the liquid layout diffs look more diff-ed than they are, I just removed the wrapper element and moved the page class into the frontmatter

### DIFF
--- a/packages/11ty/_layouts/base.11ty.js
+++ b/packages/11ty/_layouts/base.11ty.js
@@ -21,9 +21,10 @@ module.exports = function(data) {
           <div class="quire no-js" id="container">
             <div
               aria-expanded="false"
-              class="quire__secondary remove-from-epub"
+              class="quire__secondary"
               id="site-menu"
               role="contentinfo"
+              data-outputs-exclude="epub,pdf"
             >
               ${this.menu({ collections, pageData })}
             </div>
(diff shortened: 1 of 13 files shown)


packages/11ty/_layouts/base.11ty.js
```diff
@@ -21,18 +21,17 @@ module.exports = function(data) {
           <div class="quire no-js" id="container">
             <div
               aria-expanded="false"
-              class="quire__secondary"
+              class="quire__secondary remove-from-epub"
               id="site-menu"
               role="contentinfo"
```
hbalenda, line 26:
```suggestion
              class="quire__secondary"
              id="site-menu"
              role="contentinfo"
              data-outputs-exclude="epub,pdf"
```

mphstudios (approved):

PR #404, opened 2022-07-21
Feature: Render markup for output type

hbalenda:
Changes:
- Include `renderOutputs` eleventy javascript function
- Update Outputs plugin `README` with how to conditionally render components for different output types

### DIFF
--- a/packages/11ty/_plugins/transforms/outputs/README.md
+++ b/packages/11ty/_plugins/transforms/outputs/README.md
@@ -18,7 +18,7 @@ An image that will be included in `pdf` and `epub`, but excluded from `html`:
 ```
 
 ### Conditional Rendering
-For components that should render different markup for different output formats, a shortcode directory can include modules for each output type: `html.js`, `epub.js`, `pdf.js`. If the output for `epub` and `pdf` are the same, you can create a `print.js` module, which will be used to generate the markup for both `epub` and `pdf`. The shortcode should then also include an `index.js` that calls the `renderOutputs` function (see example below).
+For components that should render different markup for different output formats, a shortcode directory can include modules for each output format: `html.js`, `epub.js`, `pdf.js`. If the output for `epub` and `pdf` are the same, you can create a `print.js` module, which will be used to generate the markup for both `epub` and `pdf`. The shortcode should then also include an `index.js` that calls the `renderOutputs` function (see example below).
 
 Example directory structures for a `video` shortcode with conditional rendering:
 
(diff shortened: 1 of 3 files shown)


packages/11ty/_plugins/transforms/outputs/README.md
```diff
@@ -12,8 +12,43 @@ An image that will be included in `html` and `pdf` output, but excluded from `ep
...
 ```
 
-
 An image that will be included in `pdf` and `epub`, but excluded from `html`:
 ```html
 <img src="image.jpg" data-outputs-exclude="html" />
 ```
+
+### Conditional Rendering
+For components that should render different markup for different output formats, a shortcode directory can include modules for each output type: `html.js`, `epub.js`, `pdf.js`. If the output for `epub` and `pdf` are the same, you can create a `print.js` module, which will be used to generate the markup for both `epub` and `pdf`. The shortcode should then also include an `index.js` that calls the `renderOutputs` function (see example below).
```
mphstudios, line 21:
```suggestion
For components that should render different markup for different output formats, a shortcode directory can include modules for each output format: `html.js`, `epub.js`, `pdf.js`. If the output for `epub` and `pdf` are the same, you can create a `print.js` module, which will be used to generate the markup for both `epub` and `pdf`. The shortcode should then also include an `index.js` that calls the `renderOutputs` function (see example below).
```

packages/11ty/_plugins/transforms/outputs/render.js
```diff
@@ -0,0 +1,38 @@
...
+      : filePath
+  })
+
+  const renderFns = await Promise.all(
+    filePaths.map((filePath) => {
+      const init = require(filePath)
+      return init(eleventyConfig)
+    })
+  )
+  const content = renderFns.map((renderFn, index) => {
```
mphstudios, line 26:
```suggestion

  const content = renderFns.map((renderFn, index) => {
```

packages/11ty/_plugins/transforms/outputs/render.js
```diff
@@ -0,0 +1,38 @@
...
+    const fragment = JSDOM.fragment(renderFn(params))
+    for (child of fragment.children) {
+      const fileName = path.parse(filePaths[index]).name
+      const output = fileName === 'print'
+        ? 'epub,pdf'
+        : fileName
+      child.setAttribute('data-outputs-include', output)
+    }
+    return fragment.firstChild.outerHTML
+  })
```
mphstudios, line 36:
```suggestion
  })
  
```

mphstudios (approved):

PR #410, opened 2022-07-22
Feature: Page IDs

hbalenda:
Changes:
- Adds page ids to `<main>` element with value of slugified page url
- Updates PDF Table of Contents links to use slugified page url


mphstudios (commented):
@hbalenda we might also update the `transformRelativeLinks` method in the PDF transform to use the slugified url.

mphstudios (approved):

PR #414, opened 2022-07-23
Feature: Figure UIDs

hbalenda:
Changes:
- Adds uid generator to `~lib`
- Adds uid to all data with an `id` property
- Uses `figure.uid` in figure id attribute, figure reference anchor tags, and modal

Fixes:
- Redundant figure id properties for figures used multiple times in one publication

### DIFF
--- a/packages/11ty/_lib/uid/index.js
+++ b/packages/11ty/_lib/uid/index.js
@@ -1,0 +1,38 @@
+function* UidGenerator() {
+  let i = 0
+  while (true) yield i++
+}
+const generator = UidGenerator()
+const uid = () => generator.next().value
+
+/**
+ * Adds a `uid` property to all objects containing an `id` property
+ * @param {Object|Array} data
+ */
+const addUids = function(data) {
+  if (!data) return
+  if (data.hasOwnProperty('id')) {
+    data.uid = `uid-${uid()}`
+    return data
+  }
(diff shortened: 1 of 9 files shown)


packages/11ty/_lib/uid/index.js
```diff
@@ -0,0 +1,38 @@
+function* UidGenerator() {
+  let i = 0
+  while (true) yield i++
+}
```
mphstudios, line 4: @hbalenda though a simply incrementer will work under most conditions it is not guaranteed to be unique. We might instead emplement a `uid` generator using `Math.random()`, `Date.now()`, a combination of the two, or a `Uuid` library.

hbalenda:
@mphstudios why wouldn't it be unique? i don't think we need a uuid, and this would generate a unique number each time it's called in a build process, but... i do realize now that adding the uid to the global data won't work, and this task is a bigger issue than i anticipated 🤔

packages/11ty/_lib/uid/index.js
mphstudios: For reference, [Uuid](https://github.com/uuidjs/uuid) is widely used, however [ShortUniqueId](https://github.com/simplyhexagonal/short-unique-id) is perhaps better suited to this requirement.

PR #417, opened 2022-07-26
Fix/dev 12644 redundant ids

hbalenda:
Changes:
- Refactors `pluralize` to handle irregular plurals (ended up not needing this, but it's still nice)
- Move pluralize out of eleventy filters into `~lib`
- Log error if duplicate ids are found in yaml data


mphstudios (approved):
@hbalenda when we refactor the cli messages it may make sense to look into using a localization library which handles localized strings and pluralization though there is no requirement for that feature.

PR #418, opened 2022-07-26
Fix: Conditional rendering bugs

hbalenda:
Fixes two issues I noticed while testing a different issue:
- Include `prettier` error in error messaging
- Render all root elements of multi-root shortcode

### DIFF
--- a/packages/11ty/_plugins/transforms/outputs/render.js
+++ b/packages/11ty/_plugins/transforms/outputs/render.js
@@ -24,17 +24,14 @@ module.exports = async function (eleventyConfig, dir, params) {
     })
   )
 
-  const content = renderFns.map((renderFn, index) => {
+  const content = renderFns.flatMap((renderFn, index) => {
     const fragment = JSDOM.fragment(renderFn(params))
-    for (child of fragment.children) {
+    return [...fragment.children].map((child) => {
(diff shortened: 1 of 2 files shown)


packages/11ty/_plugins/transforms/outputs/render.js
```diff
@@ -24,17 +24,14 @@ module.exports = async function (eleventyConfig, dir, params) {
...
+  const content = renderFns.flatMap((renderFn, index) => {
     const fragment = JSDOM.fragment(renderFn(params))
-    for (child of fragment.children) {
+    return [...fragment.children].map((child) => {
       const fileName = path.parse(filePaths[index]).name
-      const output = fileName === 'print'
-        ? 'epub,pdf'
-        : fileName
+      const output = fileName === 'print' ? 'epub,pdf' : fileName
       child.setAttribute('data-outputs-include', output)
```
mphstudios, line 32:
```suggestion
      const outputs = fileName === 'print' ? 'epub,pdf' : fileName
      child.setAttribute('data-outputs-include', outputs)
```

mphstudios (changes requested):
See inline suggestions to align attribute and variable names.

mphstudios (approved):

PR #421, opened 2022-07-27
DEV-12759: Refactor license icons to use individual SVGs

anderspollack:
Previously, all license icon paths were defined in a big string literal, each defined in a `<symbol id="...">` with some confusing combination logic. 

Now, each icon is inside its own `.svg` file. Each `<symbol id="">` tag is now written from the contents of the SVG.

The full license icon in the menu now combines individual icons inside the license link, depending on the license abbreviation. For example:

```yaml
license:
  abbreviation: CC BY-NC-SA
```

yields four SVGs: `#cc`, `#by`, `#nc`, `#sa`


packages/11ty/_includes/components/icons-cc/index.js
```diff
@@ -0,0 +1,30 @@
+const fs = require('fs')
+const path = require('path')
+const ccIcons = fs
+  .readdirSync(path.join(__dirname, 'icons'))
+  .map((filename) => {
+    const id = path.basename(filename, '.svg')
+    return fs.readFileSync(
+      path.join(__dirname, 'icons', filename),
+      { encoding: 'utf8' }
+    ).replace('<svg', `<symbol id="${id}"`).replace('</svg', '</symbol')
```
anderspollack, line 10: this double replace to change both opening and closing tags from `svg` to `symbol` is a little eh

mphstudios (approved):
Nice.

PR #423, opened 2022-07-27
Feature: Update styles and build tasks for print

fchasen:
Updates the print css styles to work with Paged.js and Prince XML:

- Updates the `build:pdf` task to use an installed version of Pagedjs-CLI for generating PDFs
- Adds a `build:prince` task to generate a PDF using PrinceXML 
- Reorganize `@page` styles
- Fix min-height applied to print css
- Hides lightbox for print


packages/11ty/package.json
```diff
@@ -6,8 +6,9 @@
   "main": ".eleventy.js",
   "scripts": {
     "build": "ELEVENTY_ENV=production eleventy",
-    "build:pdf": "npx pagedjs-cli@latest ${ELEVENTY_OUTPUT_DIR:-_site}/pdf.html",
-    "clean": "del ${ELEVENTY_OUTPUT_DIR:-_site} public/pdf.html",
+    "build:pdf": "pagedjs-cli public/pdf.html --outline-tags 'h1' -o paged.pdf",
+    "build:prince": "prince public/pdf.html --output prince.pdf",
+    "clean": "del ${ELEVENTY_OUTPUT_DIR:-_site} public",
```
mphstudios, line 11: In case a user puts something else in the `public` directory we can use a glob pattern here
```suggestion
    "clean": "del ${ELEVENTY_OUTPUT_DIR:-_site} public/*.pdf",
```

fchasen: Ah great, updated. (I also added a CSS glob)

fchasen:
@mphstudios A note that the URLs all point to `/_assets/` so don't work correctly in the public folder or on a local file systems yet.

I've been moving the pdf html and css manually to the base of the `_site` folder and running a local server to generate the pdfs.

mphstudios:
> @mphstudios A note that the URLs all point to `/_assets/` so don't work correctly in the public folder or on a local file systems yet.

It may be that we need an additional transform function that corrects relative URLs for the `public/*.pdf` files.

fchasen:
https://github.com/thegetty/quire/blob/fchasen/feature/11ty/packages/11ty/_plugins/transforms/pdf/transform.js#L29-L57 was my attempt at those transforms, but it feels a bit fragile (and doesn't change the css files).

mphstudios (approved):

PR #426, opened 2022-08-01
Add ability to crop during image process

hbalenda:
- Refactors `tileImage` and `createImage` to accept `figure` data as a parameter instead of `path` so all image process methods have the same signature (and gives us access to `region`)
- Crops the print and thumbnail images if `figure.region` is specified in `figures.yaml`

### DIFF
--- a/packages/11ty/_plugins/iiif/process/createImage.js
+++ b/packages/11ty/_plugins/iiif/process/createImage.js
@@ -15,24 +15,25 @@ module.exports = (eleventyConfig) => {
   /**
    * Creates an image in the output directory with the name `${name}${ext}`
    *
-   * @param  {String} input Input file path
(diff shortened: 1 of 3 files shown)


packages/11ty/_plugins/iiif/process/createImage.js
```diff
@@ -15,32 +15,43 @@ module.exports = (eleventyConfig) => {
...
       if (debug) {
-        info(`Created ${filename}`)
+        info(`Created ${src}`)
       }
       try {
-        return await sharp(inputPath)
-          .resize(transformation.resize)
+        return await service
+          .crop(region)
+          .resize(resize)
```
hbalenda, line 50: @mphstudios I could use some feedback on this pattern - not sure this is exactly right

mphstudios (approved):

PR #430, opened 2022-08-03
Refactor/DEV-12651: Include missing PDF pages "title" and "half-title" and "copyright"

hbalenda:
Changes:
- Adds markdown templates that use the base layout and `outputs: pdf` for the `pdf-title`, `pdf-half-title`, and `copyright` pages. These were previously included in the `cover` layout and only rendered in pdf, but making them separate templates makes it easier to understand and edit these pages.
- Updates `order` properties to accommodate new pages

I'm not totally sure about the blank pages that are supposed to be included between these and the cover page, the current prince output includes the blank pages but paged.js does not, might be something to ask Fred.

### DIFF
--- a/packages/11ty/content/pdf-copyright.md
+++ b/packages/11ty/content/pdf-copyright.md
@@ -1,0 +1,9 @@
+---
+class: backmatter
+layout: base.11ty.js
+order: 4
+outputs:
+  - pdf
+title: Copyright
+---
+
--- a/packages/11ty/content/pdf-copyright.md
+++ b/packages/11ty/content/pdf-copyright.md
@@ -7,3 +7,8 @@ outputs:
 title: Copyright
 ---
 
+{% copyright %}
+
+{% if publication.identifier.isbn %}
+ISBN: {{ publication.identifier.isbn }}
+{% endif %}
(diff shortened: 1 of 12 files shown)


packages/11ty/content/pdf-copyright.md
```diff
@@ -0,0 +1,9 @@
+---
+class: backmatter
+layout: base.11ty.js
+order: 4
+outputs:
+  - pdf
+title: Copyright
+---
+
```
hbalenda, line 9: @geealbers Two questions for you:
1) Are the title, half-title, and copyright pages also included in EPUB?
2) What example content should the starter copyright page include here?

anderspollack (approved):

packages/11ty/content/pdf-copyright.md
geealbers: @hbalenda to #1, yes, those should appear in the EPUB as well. 

For #2, To start with the most basic, I would include the copyright info (as it appears in the online sidebar menu though I'd remove the publisher logo[s]) and then also include the ISBN if one is provided:

```
{% copyright %}

{% if publication.identifier.isbn %}
ISBN: {{ publication.identifier.isbn }}
{% endif %}
```

PR #433, opened 2022-08-04
Render figure images and video poster images with a relative `src` path in print output

anderspollack:
In print output, the leading `/` in front of asset paths has been stripped out, e.g. `_assets/figures/evans-graveyard.jpg`)


hbalenda:
Is there a way to make this a part of the `pdf` and `epub` transform? I'm worried that this will be difficult to maintain if we add additional components that use assets in print.

PR #442, opened 2022-08-11
Refactor/DEV-12907: IIIF global data

hbalenda:
While starting to work on annotations, I found the code (that I wrote) to add IIIF data to figures confusing, and spent some time cleaning it up. 

Changes:
- All figures now have the same `iiif` properties in `figure.iiif`, whether or not they are defined. 
- Divvies up helper methods to compute or fetch IIIF properties into `_plugins/iiif/helpers/`
- Moves the `figureIIIF` method out of eleventy filters and into `_plugins/iiif/helpers/`

Likely will continue to iterate on this as I work more on annotations, but was trying to keep this scope of work from creeping.


packages/11ty/_plugins/iiif/helpers/get-image-service.js
```diff
@@ -0,0 +1,17 @@
...
+ * @return {String}            Path to figure's `info.json`
+ */
+module.exports = (eleventyConfig, figure) => {
+  const { imageServiceDirectory, outputDir } = eleventyConfig.globalData.iiifConfig
+  const { src } = figure
+  if (!src) return
+  const { name } = path.parse(src)
+  return src.startsWith('http')
+    ? src
+    : path.join('/', outputDir, name, imageServiceDirectory, 'info.json')
```
anderspollack, line 16: This all looks good to merge and makes sense to me -- I do wonder if this ^ thing we do in like 4-5 different places where we check if a string starts with `http` and conditionally render a either the original `src` or some different relative path is worth abstracting into some kind of utility (separate task)

anderspollack (approved):

packages/11ty/_plugins/iiif/helpers/get-image-service.js
hbalenda: Wrote up a ticket fer it: https://jira.getty.edu/browse/DEV-12941

PR #458, opened 2022-08-23
Fix lightbox slides not advancing / not opening from specific figure

anderspollack:
- Ensure lightbox slides shortcode does not render anything when mediaType is unsupported
- Provide a default value for video poster param in print output
- Ensure lightbox slides provide a unique figure id value for `data-lightbox-slide` to manage slide navigation

### DIFF
--- a/packages/11ty/_includes/components/figure/video/print.js
+++ b/packages/11ty/_includes/components/figure/video/print.js
@@ -2,6 +2,8 @@ const { html } = require('~lib/common-tags')
 const chalkFactory = require('~lib/chalk')
 const path = require('path')
 
+const { warn } = chalkFactory('Figure Video')
+
 /**
  * Renders an image instead of a video player
  *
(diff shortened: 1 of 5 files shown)


packages/11ty/_includes/components/figure/video/print.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(eleventyConfig) {
 
   const { figureLabelLocation, imageDir } = eleventyConfig.globalData.config.params
 
-  return function({ aspect_ratio: aspectRatio, caption, credit, id, label, mediaType, poster}) {
+  return function({ aspect_ratio: aspectRatio, caption, credit, id, label, mediaType, poster=''}) {
```
anderspollack, line 18: I was testing out static video figures to ensure the media type exclusion for lightbox slides was working and noticed this would throw errors if a static video figure did NOT have a poster

mphstudios: A warning when a video does not have a `poster` image defined would be helpful.

packages/11ty/_includes/components/lightbox/slides.js
```diff
@@ -47,7 +48,7 @@ module.exports = function(eleventyConfig) {
       return html`
         <div
           class="q-lightbox-slides__slide"
-          data-lightbox-slide
+          data-lightbox-slide="${id}"
```
mphstudios, line 51: Does it make sense to name this verbosely, `data-lightbox-slide-id`?

PR #459, opened 2022-08-23
Use chalk for better error handling in figure video component

anderspollack:


mphstudios (commented):
@anderspollack if we have access to the input path it would help users fix the errors to add the input template to the error messages.

mphstudios (approved):

PR #461, opened 2022-08-24
Feature/DEV-12836: Print entry figures

hbalenda:
Changes:
- Adds page object to `renderOutputs` function (not used in this PR, but might need in the future)
- Removes oldie print handling from `label` shortcode and returns empty string if label is undefined
- Renders print output for entry figures using `data-outputs-include` 

Noticed there are extra blank pages between entry figures in print now, and I'm not sure why.

### DIFF
--- a/packages/11ty/_layouts/entry.liquid
+++ b/packages/11ty/_layouts/entry.liquid
@@ -32,7 +32,7 @@ Entry content, including entry image and tombstone data
   <div class="quire-entry__image-wrap" data-outputs-include="epub,pdf">
     {% if image %}
       {% comment %} 
-        To do: add support for images without ids defined in front matter
+        @TODO: add support for images without ids defined in front matter
         {% figure image %}
       {% endcomment %}
     {% elsif pageObjects %}
--- a/packages/11ty/_layouts/entry.liquid
+++ b/packages/11ty/_layouts/entry.liquid
@@ -5,33 +5,22 @@ description: Entry layout. This template is intended for use in catalogue-style 
 ---
 
 {% assign entryPageSideBySideLayout = config.params.entryPageSideBySideLayout %}
-{% assign epub = config.params.epub %}
-{% assign imageDir = config.params.imageDir %}
-{% assign pdf = config.params.pdf %}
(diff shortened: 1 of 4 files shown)


packages/11ty/_layouts/entry.liquid
```diff
@@ -5,33 +5,22 @@ description: Entry layout. This template is intended for use in catalogue-style
...
+<div {% if entryPageSideBySideLayout == true or presentation == 'side-by-side' %} class="side-by-side" {% endif %}>
 
   {% comment %} Full-width entry image header {% endcomment %}
-  <div class="quire-entry__image-wrap">
+  <div class="quire-entry__image-wrap" data-outputs-include="html">
     <div class="quire-entry__lightbox">
       {% if image %}
-        {% comment %} TODO add `lightbox` when a single `image` is included in front matter {% endcomment %}
+      {% comment %} 
+        To do: add support for images without ids defined in front matter
```
mphstudios, line 20:
```suggestion
        @TODO: add support for images without ids defined in front matter
```

packages/11ty/_layouts/entry.liquid
```diff
@@ -40,19 +29,32 @@ Entry content, including entry image and tombstone data
     </div>
   </div>
 
+  <div class="quire-entry__image-wrap" data-outputs-include="epub,pdf">
+    {% if image %}
+      {% comment %} 
+        To do: add support for images without ids defined in front matter
```
mphstudios, line 35:
```suggestion
        @TODO: add support for images without ids defined in front matter
```

packages/11ty/_plugins/transforms/outputs/render.js
```diff
@@ -21,7 +21,7 @@ module.exports = function (eleventyConfig, dir, params) {
 
   const content = filePaths.flatMap((filePath, index) => {
     const init = require(filePath)
-    const renderFn = init(eleventyConfig)
+    const renderFn = init(eleventyConfig, { page })
```
mphstudios, line 24: This may be helpful for logging messages.

mphstudios (approved):

PR #462, opened 2022-08-24
DEV-13012: Remove `figureLabelLocation` config option; always render labels in figure captions; clean up markup and unused styles

anderspollack:
- Remove all uses of `figureLabelLocation`
- Remove `--on-top` modifier styles
- Remove `figureLabelLocation` option from `config.yaml`
- Clean up table figure markup
- Remove use of `figureLabelLocation` from placeholder figure

### DIFF
--- a/packages/11ty/_includes/components/figure/modal-link.js
+++ b/packages/11ty/_includes/components/figure/modal-link.js
@@ -6,8 +6,8 @@ module.exports = function (eleventyConfig) {
   return function({ content, id }) {
     return figureModal
       ? html`<a
-          href="#${id}"
           class="q-figure__modal-link"
+          href="#${id}"
         >
           ${content}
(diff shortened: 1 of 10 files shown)


packages/11ty/_includes/components/figure/modal-link.js
```diff
@@ -6,8 +6,8 @@ module.exports = function (eleventyConfig) {
   return function({ content, id }) {
     return figureModal
       ? html`<a
-          href="#${id}"
           class="q-figure__modal-link"
+          href="#${id}"
         >
```
mphstudios, line 11: Could this be on one line?
```
html`<a class="q-figure__modal-link" href="#${id}">
```

mphstudios (approved):

PR #465, opened 2022-08-24
Feature/DEV-12650: Refactor section navigation data

hbalenda:
Replaces many lines of code with the `eleventyNavigation` method filtered by `key`. See: https://www.11ty.dev/docs/plugins/navigation/#example-get-just-one-branch


mphstudios (approved):
I like how this refactoring removes code in favor of the navigation plugin.

PR #470, opened 2022-08-31
Refactor IIIF plugin

hbalenda:
Changes:
- Refactor IIIF plugin as figure plugin that iterates over figures and initializes a `FigureFactory` instance for each that creates and processes figures.
- Refactor image tiler as `Tiler` class


packages/11ty/_plugins/figures/figure.js
```diff
@@ -0,0 +1,43 @@
...
+const ManifestWriter = require('./iiif/manifest/writer')
+
+module.exports = class Figure {
+  constructor(eleventyConfig, data) {
+    this.iiifConfig = eleventyConfig.globalData.iiifConfig
+    this.data = {
+      ...data,
+      isCanvas: isCanvas(data),
+      isImageService: isImageService(data),
+      printImage: getPrintImage(eleventyConfig, data)
```
mphstudios, line 14: Using the class pattern each of these might be to level properties rather than nested as properties of `this.data`.

packages/11ty/_plugins/figures/iiif/tile-image.js
```diff
@@ -0,0 +1,72 @@
...
+      error: `Image file type is not supported. Supported extensions are: ${supportedImageExtensions.join(', ')}`,
+      src
+    }
+  }
+
+  fs.ensureDirSync(outputPath)
+
+  const iiifId = path.join(
+    baseURL,
+    outputDir.split(path.sep).slice(1).join(path.sep),
```
mphstudios, line 49: As this part of the id will be the same for every image it might be a configuration property of a tiler instance or a tiling module that wraps `sharp`.

packages/11ty/_plugins/figures/index.js
```diff
@@ -0,0 +1,17 @@
+const Figure = require('./figure')
+const iiifConfig = require('./config')
+
+module.exports = function (eleventyConfig, options = {}) {
+  eleventyConfig.addGlobalData('iiifConfig', iiifConfig(eleventyConfig))
```
mphstudios, line 5: Since the `config` module exports a function that expects the `eleventyConfig` as an argument updating the global data with with a `iiif` or `iiifConfig` property may be a concern of the `config` module.
```suggestion
  config(eleventyConfig)
```

packages/11ty/_plugins/figures/index.js
```diff
@@ -0,0 +1,17 @@
...
+  eleventyConfig.on('eleventy.before', async () => {
+    let { figure_list: figureList } = eleventyConfig.globalData.figures
+    figureList = await Promise.all(
+      figureList.map((data) => {
+        const figure = new Figure(eleventyConfig, data)
+        return figure.create()
+      })
+    )
+    Object.assign(eleventyConfig.globalData.figures.figure_list, figureList)
+  })
```
mphstudios, line 16: Is renaming `figures_list` as `figures` confusing here? It would be nice to be able to eventually drop the additional `*_list` keys from the yaml arrays.
```suggestion
  eleventyConfig.on('eleventy.before', async () => {
    let { figure_list: figures } = eleventyConfig.globalData.figures
    figures = await Promise.all(
      figures((data) => {
        const figure = new Figure(eleventyConfig, data)
        return figure.create()
      })
    )
    Object.assign(eleventyConfig.globalData.figures.figure_list, figures)
  })
```

packages/11ty/_plugins/figures/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+module.exports = function (eleventyConfig, options = {}) {
+  eleventyConfig.addGlobalData('iiifConfig', iiifConfig(eleventyConfig))
+
+  eleventyConfig.on('eleventy.before', async () => {
+    let { figure_list: figureList } = eleventyConfig.globalData.figures
+    figureList = await Promise.all(
+      figureList.map((data) => {
+        const figure = new Figure(eleventyConfig, data)
+        return figure.create()
```
mphstudios, line 12: Making the statements in the `create` method part of the `Figure.constructor` would be more idiomatic however they may be asynchronous in which case refactoring the the `Figure` class as a FigureFactory class that returns a single method may another approach, for example
```suggestion
        const factory = new FigureFactory(eleventyConfig)
        return factory(data) 
```
the `factory` could also simply be a module with the a `create` method
```
        const factory = new FigureFactory(eleventyConfig)
        return factory.create(data) 
```

packages/11ty/_plugins/figures/transform.js
```diff
@@ -9,24 +9,26 @@ const { info, error } = chalkFactory('plugins:iiif:createImage')
...
+
+  const transform = async (transformation) => {
+    /**
+     * Creates an image in the output directory with the name `${name}${ext}`
+     *
+     * @param  {Object} figure Figure entry data from `figures.yaml`
+     * @param  {Object} transformation A transformation item from `iiif/config.js#imageTransformations`
+     * @property  {String} name The name of the file
+     * @property  {Object} resize Resize options for `sharp`
+     */
```
mphstudios, line 30: Should this be the following?
```suggestion
  /**
   * Creates a `sharp/transform` that writes the image file in the output directory
   * Nota bene: this `transform` is distinct form the `11ty/transform`
   *
   * @param  {Object} transformation A transformation item from `iiif/config.js#imageTransformations`
   * @property  {String} name The name of the file
   * @property  {Object} resize Resize options for `sharp`
   */
  const transform = async (transformation) => {
```

packages/11ty/_plugins/figures/transform.js
```diff
@@ -64,4 +66,8 @@ module.exports = (eleventyConfig) => {
       }
     }
   }
+
+  return Promise.all(imageTransformations.map((transformation) => {
+    return transform(transformation)
+  }))
```
mphstudios, line 72: Perhaps succinctly
```suggestion
  return Promise.all(transformations.map((t) => transform(t)))
```

packages/11ty/content/_computed/eleventyComputed.js
```diff
@@ -77,7 +77,7 @@ module.exports = {
    */
   pageFigures: ({ figure, figures }) => {
     if (!figure || !figure.length) return
-    return figure.map((figure) => figures.figure_list.find((item) => item.id === figure.id))
+    return figure.map((figure) => figures.find((item) => item.id === figure.id))
```
mphstudios, line 80:
```suggestion
    return figure.map((figure) => figures.find(({ id }) => id === figure.id))
```

packages/11ty/_plugins/figures/iiif/tile-image.js
```diff
@@ -0,0 +1,72 @@
...
+  )
+
+  try {
+    if (debug) {
+      console.log(`tileImage`, inputPath)
+    }
+    const response = await sharp(inputPath)
+      .toFormat(format.output.replace('.', ''))
+      .tile({
+        id: iiifId,
```
mphstudios, line 60: Renaming `iiifId` to `filePath` may reveal more clearly that we are using the path as the id.
```suggestion
        id: filePath,
```

packages/11ty/_plugins/figures/index.js
```diff
@@ -10,8 +10,8 @@ module.exports = function (eleventyConfig, options = {}) {
     let { figure_list: figureList } = eleventyConfig.globalData.figures
     figureList = await Promise.all(
       figureList.map((data) => {
-        const figure = new Figure(eleventyConfig, data)
-        return figure.create()
+        const factory = new FigureFactory(eleventyConfig, data)
+        return factory.create()
       })
```
mphstudios, line 15: Instead of creating a `factory` for every figure the class constructor can be used to configure the factory with the `eleventyConfig` and then pass `data` for each figure individually to the `create` method.

packages/11ty/_plugins/figures/annotation-factory.js
```diff
@@ -0,0 +1,58 @@
...
+
+  get filepath() {
+    return this.figure.preset === "zoom"
+      ? [
+          this.iiifConfig.outputDir,
+          this.data.id,
+          this.iiifConfig.imageServiceDirectory,
+        ].join('/')
+      : [this.iiifConfig.inputDir, this.data.src].join('/');
+  }
```
mphstudios, line 22: A JSDoc here to state that `preset` zoom serves the output of the iiif tiling processing output otherwise use the assets input directory.

packages/11ty/_plugins/figures/annotation-factory.js
```diff
@@ -0,0 +1,58 @@
...
+      case !!id:
+        return id
+      case !!label:
+        return label.split(' ').join('-').toLowerCase()
+        break;
+      case !!src:
+        return path.parse(src).name
+        break;
+      default:
+        error(`Error setting ID for annotation on figure "${this.figure.id}". Annotations must have a 'label' or 'src' property.`)
```
mphstudios, line 36: We might want to use the logger or throw to handle adding the "Error" prefix to logging, thoughts?
```suggestion
       error(`Unable to set an id for annotation on figure "${this.figure.id}". Annotations must have an 'id', 'label', or 'src' property.`)
```

packages/11ty/_plugins/figures/annotation-factory.js
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Quire Annotation
+ */
+module.exports = class AnnotationFactory {
+  constructor(iiifConfig) {
+    this.iiifConfig = iiifConfig
+  }
+
+  create(figure, data) {
```
mphstudios, line 14: JSDoc with a `typedef` for `figure` and possible the shape of `data` may be helpful.

packages/11ty/_plugins/figures/annotation-factory.js
```diff
@@ -0,0 +1,58 @@
+const chalkFactory = require('~lib/chalk')
+const path = require('path')
+const titleCase = require('~plugins/filters/titleCase')
+const { error } = chalkFactory('Figure Processing:IIIF:Annotations')
+
+/**
+ * Quire Annotation
+ */
```
mphstudios, line 8: Document here how a Quire Annotation (the Object returned by the factory) corresponds to a Annotation per the IIIF/W3C Annotation spec.

packages/11ty/_plugins/figures/annotation-factory.js
```diff
@@ -0,0 +1,58 @@
...
+module.exports = class AnnotationFactory {
+  constructor(iiifConfig) {
+    this.iiifConfig = iiifConfig
+  }
+
+  create(figure, data) {
+    const id = () => {
```
mphstudios, line 15: Add JSDoc for function that describes the logic to determine the `id`

packages/11ty/_plugins/figures/figure-factory.js
```diff
@@ -0,0 +1,95 @@
+const AnnotationFactory = require('./annotation-factory')
+const chalkFactory = require('~lib/chalk')
+const Tiler = require('./iiif/tiler')
+const transform = require('./transform')
+const Manifest = require('./iiif/manifest')
+const ManifestWriter = require('./iiif/manifest/writer')
+const logger = chalkFactory('Figure Processing')
+const { getPrintImage, isCanvas, isImageService } = require('./helpers')
```
mphstudios, line 8: Sort `require` expressions before other `const` declaration expressions.
```suggestion
const AnnotationFactory = require('./annotation-factory')
const chalkFactory = require('~lib/chalk')
const Tiler = require('./iiif/tiler')
const transform = require('./transform')
const Manifest = require('./iiif/manifest')
const ManifestWriter = require('./iiif/manifest/writer')
const { getPrintImage, isCanvas, isImageService } = require('./helpers')

const logger = chalkFactory('Figure Processing')
```

packages/11ty/_plugins/figures/iiif/manifest/index.js
```diff
@@ -26,32 +26,30 @@ module.exports = class Manifest {
...
+        return this.createAnnotation({
+          body: this.createAnnotationBody(item),
+          id: item.id,
+          motivation: item.motivation
         })
       })
+      /**
+       * Add the "base" image as a canvas annotation
+       * @param  {[type]} this.figure.baseImage [description]
+       * @return {[type]}                       [description]
```
mphstudios, line 42: type? ... description?

packages/11ty/_plugins/figures/annotation-factory.js
hbalenda: The error message is already red, that might be enough?

mphstudios (approved):
@hbalenda this refactoring is a significant simplification of the image processing.

PR #481, opened 2022-09-08
Define pageLabelDivider to config, and wrap it in spans in the markup

geealbers:
Needed better control of the divider between the page label and the page title for _Bronze Guidelines_. And it seemed like a useful change for core Quire as well so I'm submitting it in this PR. I also took the opportunity to define the value in config.yaml rather than leave it hard coded in the templates, though I have a fallback in the templates if `pageLabelDivider` is not present.

### DIFF
--- a/packages/11ty/_includes/components/table-of-contents/item/grid.js
+++ b/packages/11ty/_includes/components/table-of-contents/item/grid.js
@@ -20,7 +20,7 @@ module.exports = function (eleventyConfig) {
   const pageTitle = eleventyConfig.getFilter('pageTitle')
   const tableOfContentsImage = eleventyConfig.getFilter('tableOfContentsImage')
   const urlFilter = eleventyConfig.getFilter('url')
-  const { imageDir } = eleventyConfig.globalData.config.params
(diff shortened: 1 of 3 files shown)


hbalenda:
Thanks, I think this is good, but I don't think you need the fallback in the template - the default value in `config.yaml` is enough and the same as how we handle `pageLabelDivider`.

packages/11ty/_includes/components/table-of-contents/item/grid.js
```diff
@@ -48,8 +48,10 @@ module.exports = function (eleventyConfig) {
      */
     const isPage = !!layout
 
+    const divider = pageContributorDivider ? pageContributorDivider : ' — '
```
mphstudios, line 51: @geealbers a default value for this could be set when `pageContributorDivider` is destructured or here with the following statement
```suggestion
    const divider = pageContributorDivider || ' — '
```

mphstudios (approved):

geealbers:
Thanks for that suggestion and fix @mphstudios (sorry I just hadn't had the chance to get back to it), but merging it in now.

PR #488, opened 2022-09-14
fix/npm scripts DEV 13148

mphstudios:


mphstudios:
Close in favor of #489

PR #492, opened 2022-09-16
Feature/DEV-13184: Select figure annotations using query paramters

hbalenda:
Changes:
- Adds `goToCanvasState` method to `javascript/application/canvas-panel.js` which scrolls to a figure, and can select annotations and/or a region
- Calls `goToCanvasState` on page load with data from query parameters

### DIFF
--- a/packages/11ty/content/_assets/javascript/application/canvas-panel.js
+++ b/packages/11ty/content/_assets/javascript/application/canvas-panel.js
@@ -59,7 +59,7 @@ const goToCanvasState = function (figureId, annotationIds=[], region) {
   /**
    * Update Canvas state
    */
-  if (region) canvasPanel.setAttribute('region', region)x
+  if (region) canvasPanel.setAttribute('region', region)
   annotationIds.forEach((id) => {
     const input = document.getElementById(id)
     if (!input) {
--- a/packages/11ty/content/_assets/javascript/application/canvas-panel.js
+++ b/packages/11ty/content/_assets/javascript/application/canvas-panel.js
@@ -1,4 +1,4 @@
-import scrollToHash from "./scroll-to-hash"
+import scrollToHash from './scroll-to-hash'
 
 /**
  * Handle annotation/choice selection with canvasPanel API
--- a/packages/11ty/content/_assets/javascript/application/canvas-panel.js
+++ b/packages/11ty/content/_assets/javascript/application/canvas-panel.js
@@ -52,8 +52,9 @@ const handleSelect = (element) => {
  * @param  {Array} annotationIds  The IIIF ids of the annotations to select
  * @param  {String} region      The canvas region
  */
-const goToCanvasState = function (figureId, annotationIds=[], region) {
+const goToCanvasState = function ({ figureId, annotationIds=[], region }) {
   const figure = document.querySelector(`#${figureId}`)
+  if (!figure) return
   const canvasPanel = figure.querySelector('canvas-panel')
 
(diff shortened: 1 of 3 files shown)


packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -1,51 +1,107 @@
...
+ * Handle annotation/choice selection with canvasPanel API
+ * {@link https://iiif-canvas-panel.netlify.app/docs/examples/handling-choice}
+ */
+const handleSelect = (element) => {
+  const figure = element.closest('.q-figure')
+  if (!figure) return
+  const canvasPanel = figure.querySelector('canvas-panel')
+  const annotationType = element.getAttribute('data-annotation-type')
+  const id = element.getAttribute('id')
+  const inputType = element.getAttribute('type')
```
mphstudios, line 13: Does it makes sense to destructure `element.attributes` for these values?

packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -1,51 +1,107 @@
...
+
+  const toggleChoice = ({ checked }) => {
+    canvasPanel.makeChoice(id, {
+      deselect: !checked,
+      deselectOthers: inputType === 'radio'
+    })
+    if (inputType !== 'checkbox') return
```
mphstudios, line 20: Whenever possible, the guard clause be the first statement, for example:
```suggestion
    if (inputType !== 'checkbox') return
    
    canvasPanel.makeChoice(id, {
      deselect: !checked,
      deselectOthers: inputType === 'radio'
    })
```

packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -1,51 +1,107 @@
...
-            canvasPanel.applyStyles(id, { opacity: Number(!!checked) })
-          }
-          toggleAnnotation(input)
-          input.addEventListener('click', ({ target }) => toggleAnnotation(target));
-          break
-        default:
-          console.warn(`Invalid annotation type ${annotationType}`)
-          break
-      }
+  if (region) canvasPanel.setAttribute('region', region)x
```
mphstudios, line 61: Is the `x` at the end of the statement a typo?
```suggestion
  if (region) canvasPanel.setAttribute('region', region)
```

packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -1,51 +1,107 @@
...
+  annotationIds.forEach((id) => {
+    const input = document.getElementById(id)
+    if (!input) {
+      console.warn(`Invalid annotation id: ${id}`)
```
mphstudios, line 65: Should we use our logger here to provide context in the console message?

packages/11ty/content/_assets/javascript/application/canvas-panel.js
hbalenda: that guard clause is only guarding the things below it, maybe there is another way to write this though

packages/11ty/content/_assets/javascript/application/canvas-panel.js
hbalenda: not sure how to do this in a way that is not very verbose, unless you have a suggestion?

mphstudios (approved):

PR #493, opened 2022-09-19
Feature: Link to annotation states

hbalenda:
Changes:
- Adds `annoref` shortcode that provides a shorthand for creating links to canvas panel annotation states.
- Adds example to `iiif-demo`
- Includes handling for `region` though it doesn't work as expected, which will require some digging, _and_ since there is no UI for resetting the viewport after setting a region programmatically I'm not sure it needs to be supported yet.

### DIFF
--- a/packages/11ty/_plugins/shortcodes/annoref.js
+++ b/packages/11ty/_plugins/shortcodes/annoref.js
@@ -1,0 +1,28 @@
+const chalkFactory = require('~lib/chalk')
+const { oneLine } = require('~lib/common-tags')
+const logger = chalkFactory(`Shortcodes:Annoref`)
+/**
+ * Style wrapped `content` as "backmatter"
+ *
+ * @param      {String}  content  content between shortcode tags
+ *
+ * @return     {boolean}  A styled HTML <div> element with the content
+ */
+module.exports = function (eleventyConfig) {
+  const getAnnotation = eleventyConfig.getFilter('getAnnotation')
+  return ({ anno='', fig, region='', text='' }) => {
+    const annoIds = anno.split(',').map((id) => id.trim())
+    const annotationUrls = annoIds.flatMap((id) => {
+      const annotation = getAnnotation(fig, id)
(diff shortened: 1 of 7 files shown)


packages/11ty/_plugins/shortcodes/annoref.js
```diff
@@ -0,0 +1,28 @@
...
+ * @return     {boolean}  A styled HTML <div> element with the content
+ */
+module.exports = function (eleventyConfig) {
+  const getAnnotation = eleventyConfig.getFilter('getAnnotation')
+  return ({ anno='', fig, region='', text='' }) => {
+    const annoIds = anno.split(',').map((id) => id.trim())
+    const annotationUrls = annoIds.flatMap((id) => {
+      const annotation = getAnnotation(fig, id)
+      return annotation ? annotation.url : []
+    })
```
mphstudios, line 18: Unless `annotation.url` is already an Array the `return` value here be `[annotation.url]` or `[]`, edited:
```suggestion
    const annotationUrls = anno.split(',').flatMap((id) => {
      const annotation = getAnnotation(fig, id.trim())
      return annotation ? [annotation.url] : []
    })
```

mphstudios (commented):
See question inline regarding `flatMap` to build the list of annotation urls.

packages/11ty/_plugins/shortcodes/annoref.js
hbalenda: I think they are equivalent, unless i'm missing a case?
```[1, null, 'gradoo'].flatMap((item) => item ? item : []) // [1, 'gradoo']```
```[1, null, 'gradoo'].flatMap((item) => item ? [item] : []) // [1, 'gradoo']```

mphstudios: Indeed, I think they are equivalent; moving `trim()` into the flatMap function may still sense?

mphstudios (approved):

PR #494, opened 2022-09-20
Fix: IIIF image directory output

hbalenda:
Allow nested IIIF image directories, and use directory name in IIIF id, so that files like `/fig0932/base.jpg` and `/fig055/base.jpg` do not need unique filenames.

Sets an `outputDir` property on the `figure` for each process to reference.

Example IIIF output directory structure:
```
- public/
  - <IIIFoutputDir>
    - <figure-id>
      - <image-filename>
      - tiles/
      - manifest.json
```


packages/11ty/_plugins/figures/figure-factory.js
```diff
@@ -1,9 +1,10 @@
 const AnnotationFactory = require('./annotation-factory')
 const chalkFactory = require('~lib/chalk')
-const Tiler = require('./iiif/tiler')
-const transform = require('./transform')
 const Manifest = require('./iiif/manifest')
 const ManifestWriter = require('./iiif/manifest/writer')
+const path = require('path')
+const Tiler = require('./iiif/tiler')
+const transform = require('./transform')
```
mphstudios, line 7: Until we have lint running on pre-commit we might manually sort based on the ESLint [sort-imports](https://eslint.org/docs/latest/rules/sort-imports) rule.

mphstudios (approved):

PR #496, opened 2022-09-20
Use unique names for inputs inline vs lightbox

hbalenda:
Ensure inline and lightbox form elements are unique (fixes lightbox checkbox select bug)
Use `annotation.id` for `input.id` instead of `url`


mphstudios (commented):
@hbalenda we might investigate an implementation that allows the `annotation-ui` component to infer the composition or render `context` otherwise naming this variable `inline=true` or `lightbox=true` may avoid the confusion of the many uses of `context`, thoughts?

mphstudios (approved):

PR #498, opened 2022-09-20
Remove environment variables and use config.yaml for production baseURL

hbalenda:
Changes:
- Remove references to .env
- Set IIIF baseURL from eleventy server config in development, and `config.baseURL` in production

### DIFF
--- a/packages/11ty/_plugins/figures/annotation-factory.js
+++ b/packages/11ty/_plugins/figures/annotation-factory.js
@@ -28,7 +28,7 @@ module.exports = class AnnotationFactory {
    * @return {Annotation}
    */
   create(figure, data) {
-    const { baseURL, imageServiceDirectory, inputDir, outputDir } = this.iiifConfig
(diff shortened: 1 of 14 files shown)


packages/11ty/_plugins/figures/annotation-factory.js
```diff
@@ -28,6 +28,7 @@ module.exports = class AnnotationFactory {
    * @return {Annotation}
    */
   create(figure, data) {
+    const { baseURL, imageServiceDirectory, inputDir, outputDir } = this.iiifConfig
```
mphstudios, line 31: We refactor `iiifConfig` to have a `dirs` property, for example:
```
iiifConfig: {
  dirs: {
    imageService: '',
    input: '',
    output: ''
  }
}
```

mphstudios (approved):
Approved though if you agree refactoring to `iiifConfig.dirs: { input, output, imageservice }` might be included in this pull-request.

PR #503, opened 2022-09-26
Update print styles and Pagedjs CLI version

fchasen:
- Updates to `pagedjs-cli@0.2.2-beta.3` for `set-string attr()` support and layout fixes
- Removes `position: fixed` rules in print media, as they are repeated on every page (confusingly... per the spec)
- Removes `column-span: all` rules as Paged.js is unable to render those currently (breaks the pagination)
- Removes `leader(dotted)` (unsupported for now) and `url` in `target-counter` to fix ToC. (`url` may work without the comma, like in https://www.w3.org/TR/css-gcpm-3/#funcdef-target-counter)
- Pull from `_site` when generating PDFs


mphstudios (commented):
@fchasen did the version scheme for the `pagedjs-cli` change; is `0.2.2-beta.3` an older version than `1.0.0-alpha.2`, which we are currently using?

fchasen:
Yes, sorry for the confusion - those alphas had the wrong version numbers for a prerelease.

mphstudios (approved):

PR #504, opened 2022-09-26
Change development asset output directory to `_site`

hbalenda:
Changes:
- When `ELEVENTY_ENV=development`, write generated static assets (for pdf and IIIF) directly to `_site/`
- Change `predev` npm script to remove `public/`
- Do *not* clean `_site` directory in development mode so that assets aren't regenerated every time the dev server is started
- Continue writing assets to `public/` for production builds


mphstudios (approved):

mphstudios (commented):
@hbalenda it would be kind to our future selves to document this configuration in the top level 11ty README and the `figures/iiif/README`.

PR #507, opened 2022-09-30
Add config to include / not include bibliographies on individual pages

geealbers:
This adds the option for users to include or not include bibliographies on individual pages through a global config value (`pageBibliography`) or an individual page-level value (`page_bibliography`). The default is `true`.

This is something that's come up in several recent Getty books, including *Bronze Guidelines*, and was also recently brought up by our community. The previous solution was to change the corresponding layout file, or to hide the biblio with custom CSS.

@thegetty/int-dev: One place that's not working fully is in `_layouts/table-of-contents.11ty.js`, I didn't know the syntax for pulling in `config` from the global data. **Can you advise?**

Otherwise, in my implementation for the liquid layouts, I used this logic to go first with the page config, and then fallback to global:

```liquid
{% if page_bibliography %}
  {% assign biblio = page_bibliography %}
{% else %}
  {% assign biblio = config.params.pageBibliography %}
{% endif %}
```
I had originally wanted to use this much more concise option:

```liquid
{% assign biblio = page_bibliography || config.params.pageBibliography %}
```
This structure was used in another instance in `cover.liquid`, but it turns out the double pipe operator is not valid in assigning variables. So, I also removed its use from `cover.liquid` (where it wasn't actually needed anyway because we don't want a fallback there).

### DIFF
--- a/packages/11ty/_layouts/cover.liquid
+++ b/packages/11ty/_layouts/cover.liquid
@@ -4,7 +4,7 @@ description: Quire publication cover page
 layout: base.11ty.js
 ---
 
-{% assign coverImage = image || publication.promo_image %}
+{% assign coverImage = image %}
 {% assign imagePath = config.params.imageDir | concat: coverImage | join: '/' %}
 
 {% if page_bibliography %}
--- a/packages/11ty/_layouts/cover.liquid
+++ b/packages/11ty/_layouts/cover.liquid
@@ -7,12 +7,6 @@ layout: base.11ty.js
 {% assign coverImage = image %}
 {% assign imagePath = config.params.imageDir | concat: coverImage | join: '/' %}
 
-{% if page_bibliography %}
-  {% assign biblio = page_bibliography %}
-{% else %}
-  {% assign biblio = config.params.pageBibliography %}
-{% endif %}
-
 <section class="quire-cover__hero hero is-fullheight">
   <div
     class="quire-cover__overlay"
@@ -53,9 +47,7 @@ layout: base.11ty.js
     <div class="container is-fluid">
       <div class="content">
         {{ content }}
-        {% if biblio != false %}
-          {% bibliography pageReferences %}
-        {% endif %}
+        {% bibliography pageReferences %}
       </div>
     </div>
   </section>
--- a/packages/11ty/_layouts/cover.liquid
+++ b/packages/11ty/_layouts/cover.liquid
@@ -7,6 +7,12 @@ layout: base.11ty.js
 {% assign coverImage = image || publication.promo_image %}
 {% assign imagePath = config.params.imageDir | concat: coverImage | join: '/' %}
 
+{% if page_bibliography %}
+  {% assign biblio = page_bibliography %}
+{% else %}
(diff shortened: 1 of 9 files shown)


packages/11ty/_layouts/cover.liquid
```diff
@@ -47,7 +53,9 @@ layout: base.11ty.js
     <div class="container is-fluid">
       <div class="content">
         {{ content }}
-        {% bibliography pageReferences %}
+        {% if biblio != false %}
```
anderspollack, line 56: any reason these can't just be
```suggestion
        {% if biblio %}
```
?

packages/11ty/_layouts/table-of-contents.11ty.js
```diff
@@ -36,6 +37,8 @@ module.exports = class TableOfContents {
       : ''
     const containerClass = presentation === 'grid' ? 'is-fullhd' : ''
 
+    const biblio = page_bibliography || config.params.pageBibliography
+
```
anderspollack, line 41: I feel like mixing liquid syntax and JS string interpolation is a little funky, I think this might read better if all the conditional logic is in JS:

```suggestion
    const bibliography = page_bibliography || config.params.pageBibliography
      ? '{% bibliography pageReferences %}'
      : ''

```

packages/11ty/_layouts/table-of-contents.11ty.js
```diff
@@ -51,7 +54,9 @@ module.exports = class TableOfContents {
           <div class="quire-contents-list ${presentation}">
             ${this.tableOfContents({ collections, currentPageUrl: page.url, key, presentation })}
             <div class="content">
+            {% if ${biblio} != false %}
               {% bibliography pageReferences %}
+            {% endif %}
```
anderspollack, line 59:
```suggestion
            ${bibliography}
```

geealbers:
Thanks @anderspollack, yes, I think that looks good and I can make those changes.

One other issue I had though was that I didn't actually know how to call `config.params.pageBibliography` within the `table-of-contents.11ty.js` file. Right now I get a `config is not defined` error. Can you advise?

geealbers:
@mphstudios, after you pointed out in our meeting that there was a lot of repetition to maintain in my original approach, it occurred to me that rather than repeating all the logic in each individual layout file that uses the `{% bibliography %}` shortcode, the logic could just go in the shortcode itself (`_plugins/shortcodes/bibliography.js`). Something like this, though might also be good to add some logging so users know when the biblios have been disabled.

```js
const biblio = page.data.page_bibliography || pageBibliography
if (biblio == false) return 
```
What do you think?

mphstudios:
@geealbers the guard clause (`if (biblio == false) return `) is good.

geealbers:
@mphstudios This is ready for another review. I had to change they way I defined the `biblio` constant, and I added some logging just for pages that include citations but were set not to display.

hbalenda:
Sorry to be jumping in this conversation late, but putting out an alternate solution: https://github.com/thegetty/quire/pull/528

hbalenda (approved):

PR #508, opened 2022-09-30
Add classes to the markup based on `tags`

geealbers:
Eleventy lets you define `tags` on Markdown pages, which can then be accessed in a group through Eleventy's `collections`. 

```liquid
{% for page in collections.vocabulary %}
  ...
{% endfor %}
```
I used this in *Bronze Guidelines* and we'll add info about it to the Quire documentation as I could see it being a useful feature for the community, and certainly us again in the future.

It was also then useful to be able to style those tagged pages in a particular way, and by adding tags in as part of our page classes, as this PR does, it eliminates the needs to add both `tags:` and a separate `class:` to every page.

### DIFF
--- a/packages/11ty/content/_computed/eleventyComputed.js
+++ b/packages/11ty/content/_computed/eleventyComputed.js
@@ -50,7 +50,11 @@ module.exports = {
    */
   pageClasses: ({ collections, class: classes, layout, page, tags }) => {
     const pageClasses = []
-    // Add classes based on tags
+    /**
+     * Add Eleventy collection tags as style classes;
+     * enables custom page styles to be defined for a collection tag.
+     * @see https://www.11ty.dev/docs/collections/
+     */
     tags ? pageClasses.push(tags) : ''
     // Add computed frontmatter and page-one classes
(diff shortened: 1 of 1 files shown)


packages/11ty/content/_computed/eleventyComputed.js
```diff
@@ -48,8 +48,10 @@ module.exports = {
   /**
    * Classes applied to <main> page element
    */
-  pageClasses: ({ collections, class: classes, layout, page }) => {
+  pageClasses: ({ collections, class: classes, layout, page, tags }) => {
     const pageClasses = []
+    // Add classes based on tags
```
mphstudios, line 53:
```suggestion
    /**
     * Add Eleventy collection tags as CSS classes 
     * allowing custom styles based on the collection
     * @see https://www.11ty.dev/docs/collections/
     */
```

mphstudios (changes requested):
@geealbers see inline comment

geealbers:
@mphstudios good idea to add more detail there, thanks! just pushed the update

mphstudios:
@geealbers per team discussion, to avoid naming collisions we are going to refactor this to use a `data` attribute, `data-tags=[]`, on the page.

PR #509, opened 2022-09-30
Porting over small changes/fixes from Bronzes

geealbers:
A handful of small changes/fixes coming out of *Bronze Guidelines*.

One includes allowing users to specify `target` in markdown links: `[lnk](url){target=_blank}`. Quire's default behavior is that internal links open in the same window, and external in a new window. I think that's good, and should be left as is, but for *Bronzes* I wanted a number of specific internal links to open in new windows and based on past community feedback I thought that others might have the same need in their projects. Certainly open for discussion though if any of you see a downside to enabling this.


mphstudios (changes requested):
See suggestions inline.

packages/11ty/.gitignore
```diff
@@ -19,6 +19,9 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# macOS
+.DS_Store
+
```
mphstudios, line 24: @hbalenda may have addressed adding `.DS_Store` to the ignore patterns in a separate pull-request.

packages/11ty/_includes/components/contributor/bio.js
```diff
@@ -47,11 +47,11 @@ module.exports = function (eleventyConfig) {
...
     const contributorPages = pages.map(({ data, url }) => {
       const { label, subtitle, title } = data
-      return `${link({
+      return `<li>${link({
         classes: ['quire-contributor__page-link'],
         name: pageTitle({ label, subtitle, title }),
         url,
-      })}`
+      })}</li>`
     })
```
mphstudios, line 55: To keep the list markup as a concern of the method, we might rewrite this as
```
const contributorPagesList = {
  const items = pages.map(({ data, url }) => {
    const { label, subtitle, title } = data
    const classes = ['quire-contributor__page-link']
    const name = pageTitle({ label, subtitle, title })
    return `<li>${link({ classes, name, url })}</li>`
  })
  return html`
    <ul>
      ${items}
    </ul>
  `
}
```
and remove the `<ul>` `</ul>` elements below on lines 67 and 69.

mphstudios (changes requested):

packages/11ty/_includes/components/contributor/bio.js
geealbers: @mphstudios, sure, though I'm getting an unexpected identifier error on that code. Seems to be expecting something different in `const contributorPagesList = { ... }`?

```
[11ty] 2. Unexpected identifier (via SyntaxError)
[11ty] Original error stack trace: /Users/galbers/GitHub/quire/packages/11ty/_includes/components/contributor/bio.js:49
[11ty]       const items = pages.map(({ data, url }) => {
[11ty]             ^^^^^
```
One solution might be to leave `contributorPages` as it was, and then add 

```js
const contributorPagesList =
    html`
      <ul>
        ${contributorPages}
     </ul>
   `
```

And then return 

```js
<div class="quire-contributor__details media-content">
  ${contributorImage}
  ${contributorBio}
  ${contributorPagesList}
</div>
```

mphstudios: @geealbers I updated/corrected my suggested rewrite of `contributorPages` to use an anonymous function, 

```
const contributorPagesList = () => {
  const items = pages.map(({ data, url }) => {
    const { label, subtitle, title } = data
    const classes = ['quire-contributor__page-link']
    const name = pageTitle({ label, subtitle, title })
    return `<li>${link({ classes, name, url })}</li>`
  })
  return html`
    <ul>
      ${items}
    </ul>
  `
}
```
then
```
<div class="quire-contributor__details media-content">
  ${contributorImage}
  ${contributorBio}
  ${contributorPagesList()}
</div>
```

PR #511, opened 2022-10-03
Enable table scrolling and refine default styles

geealbers:
Some table-related fixes that I made first in _Bronze Guidelines_:

- Add a a div wrapper and the necessary css to enable scrolling on large tables in the lightbox
- Clean up default table styles in the lightbox and for tables on the page
- Fix a series of classes that users can add in the figure shortcode to make table font size smaller (ie., `is-table-scale-7`)
- Adjusted the z-index of several UI elements so that they weren't hidden behind the table. (Note, I put them all in `q-lightbox-ui.scss` for clarity, though technically one is part of the modal, and one is part of the lightbox slides.)
- Add a couple large tables to the starter material for testing

### DIFF
--- a/packages/11ty/_includes/components/lightbox/slides.js
+++ b/packages/11ty/_includes/components/lightbox/slides.js
@@ -43,7 +43,7 @@ module.exports = function(eleventyConfig) {
           case mediaType === 'soundcloud':
             return figureAudioElement(figure)
           case mediaType === 'table':
-            return `<div class="table-wrapper">${await figureTableElement(figure)}</div>`
+            return `<div class="overflow-container">${await figureTableElement(figure)}</div>`
           case isVideo:
             return figureVideoElement(figure)
           case mediaType === 'image':
--- a/packages/11ty/_includes/components/lightbox/slides.js
+++ b/packages/11ty/_includes/components/lightbox/slides.js
@@ -43,7 +43,7 @@ module.exports = function(eleventyConfig) {
           case mediaType === 'soundcloud':
             return figureAudioElement(figure)
           case mediaType === 'table':
-            return await figureTableElement(figure)
+            return `<div class="table-wrapper">${await figureTableElement(figure)}</div>`
           case isVideo:
             return figureVideoElement(figure)
           case mediaType === 'image':
--- a/packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
+++ b/packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
@@ -44,17 +44,12 @@
       justify-content: center;
     }
 
-    &--table .table-wrapper {
-      background-color: white;
-      color: black;
-      display: flex;
-      flex-direction: column;
-      max-height: 80vh;
-      max-width: 80vw;
-      overflow: scroll;
-      padding: 16px;
+    &--table {
 
-      table,
+      table {
+        background-color: white;
+        color: black;
+      }
       caption {
         margin-bottom: 1rem;
       }
@@ -122,6 +117,16 @@
 
     img {
       object-fit: contain;
+    }
+
+    .overflow-container {
+      background-color: white;
+      color: black;
+      display: flex;
+      flex-direction: column;
+      max-height: 80vh;
+      max-width: 80vw;
+      padding: 1rem;
     }
   }
 
--- a/packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
+++ b/packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
@@ -46,6 +46,16 @@
 
     &--table {
 
+      .overflow-container {
+        background-color: white;
+        color: black;
+        display: flex;
+        flex-direction: column;
+        max-height: 80vh;
+        max-width: 80vw;
+        padding: 1rem;
+      }
+
       table {
         background-color: white;
         color: black;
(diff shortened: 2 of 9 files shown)


packages/11ty/_includes/components/lightbox/slides.js
```diff
@@ -43,7 +43,7 @@ module.exports = function(eleventyConfig) {
           case mediaType === 'soundcloud':
             return figureAudioElement(figure)
           case mediaType === 'table':
-            return await figureTableElement(figure)
+            return `<div class="table-wrapper">${await figureTableElement(figure)}</div>`
```
mphstudios, line 46: @geealbers I think that this can be done without the wrapping element

```css
table {
  display: block;
  max-height: 80vh;
  max-width: 80vw;
  overflow: auto;
}
```

Though if there are reasons to use a wrapping element I would suggest a simple `tableContainter` class, for example

```css
.overflow-container > table {
  max-height: 80vh;
  max-width: 80vw;
  overflow: auto;
}
```

@anderspollack thoughts?

anderspollack: I played around with it for a bit, and I think the wrapper is necessary--the `<table>` doesn't quite work with the rows like a generic container. but I agree that a more generalized `overflow-container` makes sense. Also currently, these table style selectors are one level more specific than they need to be. 

```
      table,
      caption {
        margin-bottom: 1rem;
      }
      table,
      th,
      td {
        border: 1px solid black;
        border-collapse: collapse;
        width: fit-content;
      }
      th,
      td {
        padding: 4px;
      }
      a {
        border-bottom: 1px dotted currentColor;
        color: currentColor;

        &:hover {
          border-bottom-style: solid;
        }
      }

```

are all nested under `.q-lightbox-slides__element--table .table-wrapper`, but I think it would be best if they were only nested under `.q-lightbox-slides__element--table` (styles specific to `.q-lightbox-slides__element--table .table-wrapper` are fine, but a more general `.overflow-container` utility class is better IMO)

geealbers:
@mphstudios and @anderspollack, one reason I used the div wrapper was because I've often had to include an additional html element along with the table markup in order to show footnote-like information below the main table. Like:

```html
<table>
...
</table>
<div class="table-notes">
...
</div>
```

This is stuff beyond what would be appropriate for the `<caption>` element or for a `<tfoot>`. So the div wrapper just made sure everything was included in the scroll. Calling it `.overflow-container` certainly makes sense though.

geealbers:
Oh @anderspollack, our comments came at the same time, but good point about the specificity of the css, I can adjust that

packages/11ty/_includes/components/lightbox/slides.js
geealbers: @anderspollack I'm cleaning this up per your comments and our discussion yesterday. I think `.overflow-container` would just have `overflow: scroll`, and the other properties here would be specific to the use of the overflow containe _for_ tables _in_ the lightbox slide? Is that right?

```scss
.overflow-container {
  overflow: scroll;
}

&--table .overflow-container {
  display: flex;
  flex-direction: column;
  max-height: 80vh;
  max-width: 80vw;
}
```

anderspollack: Yep that sounds right to me! Although I might define the 
```
.overflow-container {
  overflow: scroll;
}
```
in `content/_assets/styles/utilities.scss` with the other utility classes like `visually-hidden`

geealbers:
@anderspollack and @mphstudios, I made the changes as discussed, including establishing `overflow-container` as a general utility class, and making some of the table css a level less specific. Can one of you do a final review and I can merge it?

packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
```diff
@@ -88,6 +118,16 @@
...
+
+    .overflow-container {
+      background-color: white;
+      color: black;
+      display: flex;
+      flex-direction: column;
+      max-height: 80vh;
+      max-width: 80vw;
+      padding: 1rem;
+    }
```
anderspollack, line 130:
```suggestion
```

packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
```diff
@@ -44,6 +44,36 @@
       justify-content: center;
     }
 
+    &--table {
+
```
anderspollack, line 48:
```suggestion
      .overflow-container {
        background-color: white;
        color: black;
        display: flex;
        flex-direction: column;
        max-height: 80vh;
        max-width: 80vw;
        padding: 1rem;
      }
```

anderspollack: I would still keep the `.overflow-container` overrides nested under `&--table`, mostly because it applies background color/dimensions/text color that may only be applicable to table content (I could potentially see a different treatment for non-table overflow containers) but other than this extremely picky nit, I think this all looks good to go

packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
anderspollack: see above comment

packages/11ty/content/_assets/styles/components/q-lightbox-slides.scss
geealbers: @anderspollack thanks. I made that change. Good to merge in now?

anderspollack (approved):

PR #513, opened 2022-10-03
Add .DS_Store to .eleventyignore

hbalenda:


mphstudios:
@hbalenda according the to Eleventy docs 
> any paths listed in your project’s .gitignore file are automatically ignored—you don’t need to duplicate them to your .eleventyignore file.
so the solution seems obscure; suggesting that we add comments to the `.eleventy.js` config where the `passthroughCopy` for `_assets` is added *and* to the `.eleventyignore` about why the `.DS_Store` pattern needs to be listed again.

packages/11ty/.eleventyignore
```diff
@@ -1,2 +1,3 @@
 _site/
 public/
+**/*/.DS_Store
```
mphstudios, line 3:
```suggestion
**/*/.DS_Store
_site/
public/
```

PR #518, opened 2022-10-05
Use annoref to select a region

hbalenda:
Fixes:
- Linking to regions with the `annoref` shortcode, and log a warning if used with non-zooming figures.

Refactors `canvas-panel.js` with `update` and `selectAnnotation` methods to separate UI and canvas property concerns.

### DIFF
--- a/packages/11ty/content/_assets/javascript/application/canvas-panel.js
+++ b/packages/11ty/content/_assets/javascript/application/canvas-panel.js
@@ -26,6 +26,9 @@ const goToCanvasState = function ({ annotationIds=[], figureId, region='' }) {
   const figure = document.querySelector(`#${figureId}, [data-lightbox-slide-id="${figureId}"]`)
   if (!figure) return
   const canvasPanel = figure.querySelector('canvas-panel')
+  if (region && canvasPanel.getAttribute('preset') !== 'zoom') {
+    console.warn(`Using the "annoref" shortcode to link to a region on a figure without zoom enabled is not supported. Please set the "preset" property to "zoom" on figure id "${figureId}"`)
+  }
   const lightbox = figure.closest('q-lightbox')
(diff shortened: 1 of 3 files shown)


mphstudios (approved):

geealbers:
@hbalenda `zoom: true` still works as an alias for `preset: zoom`, correct? The warning message may need to account for this, and in documentation, we prefer `zoom: true`.

hbalenda:
@geealbers They mean different things, I may have done a poor job of explaining when we settled on `zoom: true`. We need both a way to indicate that a figure should be tiled (currently, `zoom: true`), and whether it should be displayed with zoom enabled when it's rendered inline with the text (`preset: zoom`).

geealbers:
@hbalenda Ah ha! yes, I didn't understand that before, but that makes sense.

That said, I'm very reluctant to have both `zoom: true` and `preset: zoom` in the mix for our users. At the very least I'd like to advocate for some wording change, but can I also ask, would it make sense to just automatically set `preset: zoom` on any image with annotations? Or if not, could we possibly do it client side by finding any annoref links on the page and then setting `preset: zoom` on the matching figures based on id? Just spitballing here

hbalenda:
@geealbers The default behavior is that figures have `preset: responsive` inline, and `preset: zoom` in the lightbox (entry or modal). I think we wanted to avoid having `preset: zoom` set on inline figures because of the scroll interaction, though I think users should still to be able to change that behavior if they want, so we need a figure property to set the zooming behavior on the display of a figure that is unique from the property that determines whether or not a figure is tiled. Open to suggestions, but I would advocate for something really explicit like `tile: true` and `zoom: true`, since I don't think we can conflate the two concerns without losing the ability to customize the behavior.

hbalenda:
@geealbers or better, we attempt to tile everything, and just skip over images that are too small.

PR #525, opened 2022-10-05
Register commands as subcommands of program

hbalenda:
Renames `Command.command` -> `Command.name`
Registers commands as subcommands of program

### DIFF
--- a/packages/cli/Command.js
+++ b/packages/cli/Command.js
@@ -28,7 +28,7 @@ export default class Command {
       throw new Error('Command is an *abstract* class')
     }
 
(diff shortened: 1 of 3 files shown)


packages/cli/Command.js
```diff
@@ -28,7 +28,7 @@ export default class Command {
       throw new Error('Command is an *abstract* class')
     }
 
-    this.command = definition.name
```
hbalenda, line 31: @mphstudios I thought it was confusing to have "command" being used to reference both the command object and command name, so I renamed this property `name`

PR #526, opened 2022-10-05
Set annotation targets on manifest json directly

hbalenda:
Builder only uses the default annotation target (the canvas id) and doesn't allow you to pass a target value. This change updates the target on the manifest, allowing us to create annotations that target a region of the canvas.

### DIFF
--- a/packages/11ty/_plugins/figures/iiif/manifest/index.js
+++ b/packages/11ty/_plugins/figures/iiif/manifest/index.js
@@ -145,6 +145,15 @@ module.exports = class Manifest {
     })
     try {
       this.json = builder.toPresentation3(manifest)
+      /**
+       * Note: Builder can't currently create annotations with target regions,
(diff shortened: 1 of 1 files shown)


packages/11ty/_plugins/figures/iiif/manifest/index.js
```diff
@@ -145,6 +145,15 @@ module.exports = class Manifest {
     })
     try {
       this.json = builder.toPresentation3(manifest)
+      /**
+       * Note: Builder can't currently create annotations with target regions,
+       * so set the target on each annotation directly
+       */
```
mphstudios, line 151: Should we add a link to the issue? https://github.com/stephenwf/iiif-builder/issues/6
```suggestion
      /**
       * Note: Builder can't currently create annotations with target regions,
       * so set the target on each annotation directly
       * @see @stephenwf/iiif-builder#6
       */
```

PR #528, opened 2022-10-06
Refactor bibliography config to use data cascade

hbalenda:
Putting this out there as an alternate solution - I was finding the bibliography logic hard to follow, and the ideal solution here is to leverage the data cascade, which we hadn't approached yet since it required refactoring the `config.yaml`. This refactors just the bibliography configuration, but shows that you can override any `config` property from a page by using the `config` key. For example in `essay.md` to override `displayOnPage` you would write:
```yaml
layout: essay
title: My Essay
config:
  bibliography:
    displayOnPage: true
```
Maybe more verbose, but I think it's simpler because you only have to remember the structure of one config instead of remembering when/where to use `page_bibliography` or `pageBibliography`, and this pattern could definitely be applied elsewhere.

### DIFF
--- a/packages/11ty/_layouts/bibliography.liquid
+++ b/packages/11ty/_layouts/bibliography.liquid
@@ -7,7 +7,7 @@ layout: page
 {{ content }}
 
 <div class="quire-page__content__references">
-  {% if config.params.displayBiblioShort == true %}
+  {% if config.bibliography.displayShort == true %}
     {% for entry in entries %}
       <dl>
         <dt><span id="{{ forloop.index }}">{{ entry.id | markdownify }}</span></dt>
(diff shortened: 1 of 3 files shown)


mphstudios (approved):
@hbalenda this better encapsulates the config and render for the bibliography and is easier to follow. Does it make sense to refactor `displayShort` from a boolean to a string, for example `displayFormt: 'short'` to allow multiple formats?

geealbers:
@hbalenda Yes, agreed, it's better this way. would we [eventually] be able to adopt this as a standard going forward that any value in config.yaml could be overridden at the page level by following the same structure?

For @mphstudios question about refactoring from a boolean `displayShort` to a string `displayFormat`, I don't think it's necessary as we're really just asking "Do you want to display the short form of the reference or not?" Whereas I'd imagine some future time where `displayFormat` might instead indicate the style of how the full reference is constructed (Chicago, MLA, etc.) were we to have a more nuanced/granular handling of bibliographic information. Say for example, allowing authors to import reference data from a tool like Zotero.

hbalenda:
@geealbers Definitely, we should come up with a list of what those features are. There is a task for refactoring the config data, if you have others they can be added there: https://jira.getty.edu/browse/DEV-12846

PR #546, opened 2022-10-18
Use polling for canvas.makeChoice

hbalenda:
Changes:
- Add polling module
- Use polling to check for `canvasPanel.makeChoice` definition before calling method


mphstudios (approved):
Polling is hacky when there is an event model though this is cleanly hacked.

PR #547, opened 2022-10-21
Add oxford comma to contributors list

geealbers:
Getty Pubs' standard is to use the Oxford comma ("one, two, and three" rather than "one, two and three") and so that should be Quire's standard too. I've done it here for the contributors list, which I needed for my current book projects.

Though I wondered if at some point you'd want to refactor this using [common-tags with oneLineCommaListsAnd](https://www.npmjs.com/package/common-tags#onelinecommalistsand) which is used on the `ref` shortcode. And looks like there are some options that could be called and so maybe made into config values for Quire users who hate the Oxford comma, or for i18n (changing "and" to "et" or "y").

```js
opts = {
  separator: ',', // what to separate each item with (always followed by a space)
  conjunction: 'and', // replace the last separator with this value
  serial: true // should the separator be included before the conjunction? As in the case of serial/oxford commas
}
```
Not a task that needs doing now, but I can put it in Jira if you'd like.

### DIFF
--- a/packages/11ty/_plugins/shortcodes/contributors.js
+++ b/packages/11ty/_plugins/shortcodes/contributors.js
@@ -6,7 +6,7 @@ const { error } = chalkFactory('shortcodes:contributors')
 /**
  * Contributor shortcode
  * Renders a list of contributors
- * 
+ *
  * @param  {Array|String} context Array of contributor objects OR string override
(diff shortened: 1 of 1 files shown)


hbalenda:
@geealbers Thanks, and yeah I think that's a great idea, please write up a task for refactoring contributors to use `oneLineCommaListsAnd`

PR #549, opened 2022-10-23
Feature/starter default

mphstudios:
*This starter template is a work in progress.*

### Changes

Replaces previous `quire/starter/default` content with `content` from the Eleventy project used during development.


packages/cli/src/lib/quire/init-starter.js
```diff
@@ -1,3 +1,6 @@
+import fs from 'fs-extra'
+import path from 'path'
```
mphstudios, line 2:
```suggestion
import path from 'node:path'
```

mphstudios:
Superseded by `@thegetty/quire-starter-default`

@todo sync git subtree at `starters/default` with `@thegetty/quire-starter-default`

PR #550, opened 2022-10-25
Refactor linking to region

hbalenda:
Linking to the region of a canvas panel or image service had a wonky user experience due to `zoom` being enabled

Changes:
- Moves `canvasHeight` and `canvasWidth` definition into `figure` model 
- Replaces `canvas.goToTarget` with `goToRegion` with a transition. Using this method allows us to use `preset=responsive` (no letterboxing, disabled user interaction with the figure) and modify the region programmatically.
- Adds a default region value of the full canvas dimensions to use to reset the viewport

Note:
- The `goToRegion` method isn't available on `image-service` components, so I refactored them as manifests with a single annotation (the image service). As a result, you can't use the annoref shortcode to link to a region of an externally hosted image service (hoping that would be an edge case).
- We also may need to add a button to reset the viewport

Still troubleshooting firefox bugs (properly scrolling to figure and region on load)

### DIFF
--- a/packages/11ty/_plugins/figures/annotation/index.js
+++ b/packages/11ty/_plugins/figures/annotation/index.js
@@ -48,11 +48,14 @@ module.exports = class Annotation {
     }
 
     /**
+     * Create image service for annotation image if it is a JPG and
+     * the figure has zoom enabled
+     * 
      * Note: Currently only JPG image services are supported by 
      * canvas-panel/image-service tags
      */
-    const isImageService = !!figure.isImageService && ext === '.jpg'
-
+    const isImageService = !!figure.zoom && ext === '.jpg'
+    console.log(figure)
     const info = () => {
       if (!isImageService) return
       const tileDirectory = path.join(outputDir, name, iiifConfig.dirs.imageService)
--- a/packages/11ty/_plugins/figures/annotation/index.js
+++ b/packages/11ty/_plugins/figures/annotation/index.js
@@ -55,7 +55,6 @@ module.exports = class Annotation {
      * canvas-panel/image-service tags
      */
     const isImageService = !!figure.zoom && ext === '.jpg'
-    console.log(figure)
     const info = () => {
       if (!isImageService) return
       const tileDirectory = path.join(outputDir, name, iiifConfig.dirs.imageService)
--- a/packages/11ty/_plugins/figures/figure/index.js
+++ b/packages/11ty/_plugins/figures/figure/index.js
@@ -3,6 +3,7 @@ const AnnotationFactory = require('../annotation/factory')
 const ImageProcessor = require('../image/processor')
 const Manifest = require('../iiif/manifest')
 const path = require('path')
+const sharp = require('sharp')
 const { isCanvas, isImageService } = require('../helpers')
 
 /**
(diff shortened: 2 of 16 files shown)


packages/11ty/_plugins/figures/annotation/index.js
```diff
@@ -48,11 +48,14 @@ module.exports = class Annotation {
...
+     * Create image service for annotation image if it is a JPG and
+     * the figure has zoom enabled
+     * 
      * Note: Currently only JPG image services are supported by 
      * canvas-panel/image-service tags
      */
-    const isImageService = !!figure.isImageService && ext === '.jpg'
-
+    const isImageService = !!figure.zoom && ext === '.jpg'
+    console.log(figure)
```
anderspollack, line 58:
```suggestion
```

geealbers:
@hbalenda, the only thing that I found off was that if you use an annoref to zoom into a canvas panel region and then click a checkbox or radio button, it resets the region back to the full image, when instead we'd want the region to stay as it was.

> The goToRegion method isn't available on image-service components, so I refactored them as manifests with a single annotation (the image service). As a result, you can't use the annoref shortcode to link to a region of an externally hosted image service (hoping that would be an edge case).

We talked previously about people potentially being able to use a Quire-generated IIIF manifest to view the image in another context (like to load it into Midador). Does this refactor make those manifests less standard or usable to other services?

hbalenda:
@geealbers I made the user interactions reset the viewport since otherwise it gets stuck at that region since the user can no longer zoom in or out. Does adding a 'reset' button and changing the annotation input behavior to not change the region seem like a good solution?

And changing the single-image figures from image services to manifests actually makes them more compatible with other viewers, i'm not sure most viewers support plain image services.

geealbers:
@hbalenda I see. Yes, "adding a 'reset' button and changing the annotation input behavior to not change the region" seems good. Could/would the button just appear at the end of the annotations UI list? We could maybe style it a little differently at some point but I like the idea of not adding another separate visual element, and I think that seems like a sensible place for it.

anderspollack (approved):
this works real nice

packages/11ty/_plugins/figures/figure/index.js
```diff
@@ -57,22 +59,33 @@ module.exports = class Figure {
   }
 
   /**
-   * If the `src` is an external resource
-   * @return {Boolean}
+   * Use dimensions of src or first choice as canvas dimensions
```
anderspollack, line 62: This comment is a little confusing, since it's describing how this getter is intended to be used later rather than what it does directly

hbalenda:
@geealbers - @anderspollack and I came up with a solution for the figure region 'reset' button so that it's cohesive with the rest of the figure controls. Let me know what you think.

Changes:
- The fullscreen icon would be moved from the caption to the upper left corner of the figure. The fullscreen icon will continue to be inline with the caption on figures that aren't images.
- A reset icon would be added to the upper left corner below the fullscreen icon on figures where the region can be changed with an `annoref`
- Clicking an `annoref` link that changes the region would enable the reset button
- Clicking the reset button would reset the figure region and disable the button
- Interacting with other annotation UI elements (checkbox/radio inputs) would change the annotation state but not the region

It could still be decided either way if the reset button is visually hidden or visible but disabled when it can't be interacted with.

geealbers:
@hbalenda and @anderspollack, that solution sounds good over all. The one issue is having the icons on the upper left corner of the image. I could foresee some authors/curators/users not wanting anything over the image (this is why we had the option to specify the figure label being on-top of or below the image in old Quire, which we then removed as being too complicated to maintain and so now only offered it below). Can we think of a different location for them and either go with that as the only option, or make it configurable for users (without reintroducing the complications we eliminated for the label position)?

And then as far as "visually hidden or visible but disabled" I think we'd want the reset button visually hidden until needed. Otherwise, you could imagine a publication full of disabled reset buttons that aren't ever used because the author never uses `annoref`.

hbalenda:
I'm gonna merge this for now, but we will address the UI concerns in a separate branch after more discussion.

PR #559, opened 2022-10-31
Feature: CLI preview

hbalenda:
Changes:
- Adds `cli.serve` method for `preview` command 
- Extract factory function to create eleventy CLI command

### DIFF
--- a/packages/cli/src/lib/11ty/cli.js
+++ b/packages/cli/src/lib/11ty/cli.js
@@ -1,29 +1,40 @@
-import execa from 'execa'
+import { execa } from 'execa'
 import { fileURLToPath } from 'node:url'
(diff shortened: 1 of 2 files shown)


packages/cli/src/lib/11ty/cli.js
```diff
@@ -2,32 +2,56 @@ import { execa } from 'execa'
 import { fileURLToPath } from 'node:url'
 import path from 'node:path'
 import paths from './paths.js'
+const projectRoot = path.resolve('../../packages/11ty')
```
mphstudios, line 5:
```suggestion

const projectRoot = path.resolve('../../packages/11ty')
```

mphstudios (approved):

PR #566, opened 2022-11-01
Replace config.baseURL with publication.url

hbalenda:
Separates concerns of publication data and configuration data:

Changes:
Removes `config.baseURL` and `publication.identifier.url` in favor of `publication.url`. 
Removes `config.title` in favor of `publication.title`. Not sure why these both existed, `config.title` was only being used (and incorrectly) in meta tags.

### DIFF
--- a/packages/11ty/_includes/components/citation/page.js
+++ b/packages/11ty/_includes/components/citation/page.js
@@ -42,7 +42,7 @@ module.exports = function (eleventyConfig) {
       'publisher-place': publishers[0].location,
       title: pageTitle(page.data),
       type: 'chapter',
-      url: new URL(page.url, url)
+      URL: new URL(page.url, url)
     }
   }
 }
--- a/packages/11ty/_includes/components/citation/page.js
+++ b/packages/11ty/_includes/components/citation/page.js
@@ -42,7 +42,7 @@ module.exports = function (eleventyConfig) {
       'publisher-place': publishers[0].location,
       title: pageTitle(page.data),
       type: 'chapter',
-      URL: new URL(page.url, url)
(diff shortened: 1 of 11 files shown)


packages/11ty/_includes/components/citation/page.js
```diff
@@ -43,7 +42,7 @@ module.exports = function (eleventyConfig) {
       'publisher-place': publishers[0].location,
       title: pageTitle(page.data),
       type: 'chapter',
-      URL: new URL(page.url, baseURL)
+      URL: new URL(page.url, url)
```
mphstudios, line 45: Should this be the `URL` object or its string representation?
```suggestion
      url: new URL(page.url, url).toString()
```

packages/11ty/_includes/components/head-tags/opengraph.js
```diff
@@ -10,17 +10,17 @@ module.exports = function(eleventyConfig) {
...
     const meta = [
       {
         property: 'og:title',
-        content: pageType != 'essay' ? config.title : page.title
+        content: pageType != 'essay' ? publication.title : page.title
       },
       {
         property: 'og:url',
-        content: pageType != 'essay' ? config.baseURL : permalink
+        content: new URL(page.url, url)
```
mphstudios, line 23:
```suggestion
        content: new URL(page.url, url).toString()
```

mphstudios (commented):
@hbalenda unless I have misread the code, we need to call `toString` on the `URL` instances.

hbalenda:
@mphstudios fixed. weirdly they were resolving to strings in those contexts, but stringifying them is much more explicit

PR #575, opened 2022-11-03
Fix Quire CLI `create` command bugs

anderspollack:


packages/cli/src/lib/quire/init-starter.js
```diff
@@ -1,38 +1,56 @@
...
+  fs.writeFileSync(configFilePath, configJSON)
+
   // Copy 11ty files
-  const fullRootPath = path.resolve(rootPath)
-  const eleventyPath = path.resolve(cwd(), path.join('quire', 'versions', quireVersion))
-  const eleventyFiles = fs.readdirSync(eleventyPath)
+  const fullProjectRootPath = resolve(projectRoot)
+  const eleventyPath = resolve(cwd(), join('quire', 'versions', quire11tyVersion))
+  // @TODO remove `.env` and `_site` from published `quire-11ty` NPM package
+  const eleventyIgnoreFiles = ['.env', '_site']
```
anderspollack, line 34: `.env` and `_site` got published with the public `@thegetty/quire-11ty` package, so I'm just ignoring them here until that gets updated

packages/cli/src/lib/quire/init-starter.js
```diff
@@ -1,38 +1,56 @@
...
 
   // Reinitialize project as a new git repository
-  await fs.remove(path.join(rootPath, '.git'))
+  await fs.remove(join(projectRoot, '.git'))
 
-  const starterFiles = fs.readdirSync(rootPath)
+  // don't git-add copied `node_modules`
+  const starterFiles = fs
+    .readdirSync(projectRoot)
+    .filter((filePath) => ![...eleventyIgnoreFiles, 'node_modules'].includes(filePath))
```
anderspollack, line 53: We want to copy the `node_modules` of the installed `@thegetty/quire-11ty` package (for now) into the project, but we don't want to `git add node_modules` during the initial commit. 

**Note:** the `.gitignore` in https://github.com/thegetty/quire-starter-default/ previously included all the 11ty directories (`_includes/`, `_layouts/`, `_plugins/`, `.eleventy.js`) which I commented out for now -- I am assuming this is because our eventual intention is to handle the 11ty pathing stuff in the `pacakges/cli/src/lib/11ty` module so we DON'T need to manually copy 11ty files into a created project

packages/cli/src/lib/quire/init-starter.js
mphstudios, line 34: @anderspollack `@thegetty/quire-11ty` has been updated to use the `.ignore` file when publishing and unpublished, it can be republished in 24 hours.
```suggestion
```

packages/cli/src/commands/create.js
```diff
@@ -1,17 +1,17 @@
-import { join, resolve } from 'node:path'
 import Command from '#src/Command.js'
 import { cwd } from 'node:process'
 import fs from 'fs-extra'
 import hostedGitInfo from 'hosted-git-info'
 import { initStarter } from '#src/lib/quire/init-starter.js'
 import installNpmVersion from 'install-npm-version'
 import { isEmpty } from '#helpers/is-empty.js'
+import { join } from 'node:path'
```
mphstudios, line 8:
```suggestion
import path from 'node:path'
```

packages/cli/src/lib/quire/init-starter.js
```diff
@@ -1,38 +1,56 @@
...
+    projectRoot: resolve(projectRoot),
+    version: quire11tyVersion
+  }
+  const configFilePath = join(projectRoot, 'project.json')
+  const configJSON = JSON.stringify(projectConfig, null, 2)
+  fs.writeFileSync(configFilePath, configJSON)
+
   // Copy 11ty files
-  const fullRootPath = path.resolve(rootPath)
-  const eleventyPath = path.resolve(cwd(), path.join('quire', 'versions', quireVersion))
```
mphstudios, line 23: Using methods on an imported `path` object is more explicit; we should figure out what is causing the import error for our local development environments.

mphstudios (changes requested):
@anderspollack this looks good though see comments inline and we can debug next week.

packages/cli/src/lib/quire/index.js
```diff
@@ -38,41 +109,23 @@ function getVersion() {
...
     }
   )
-
-  fs.symlink(`${INSTALL_PATH}/latest`, latest, 'dir', (error) => {
-    console.error(error)
-  })
 }
 
 /**
  * Retrieves version from quire repository `packages/11ty/package.json`
```
mphstudios, line 122:
```suggestion
 * Retrieve the `latest` published version of the `quire-11ty` package
```

packages/cli/src/lib/quire/index.js
```diff
@@ -38,41 +109,23 @@ function getVersion() {
...
  */
 async function latest() {
-  const latestQuirePackageJsonUrl = hostedGitInfo
-    .fromUrl('git@github.com:thegetty/quire.git')
-    .file('packages/11ty/package.json')
-  const latestQuirePackageJsonRequest = await fetch(latestQuirePackageJsonUrl)
-  const latestQuirePackageJson = await latestQuirePackageJsonRequest.json()
-  const { version } = latestQuirePackageJson
-  return version
+  const { stdout: quireVersion } = await execa('npm', ['view', PACKAGE_NAME, 'version'])
```
mphstudios, line 127: We may need to revisit this to handle `stderr` as well.

mphstudios (approved):

PR #585, opened 2022-11-11
Fix pathing issues in `quire new` command and `lib/quire` functions

anderspollack:
This PR:
- Fixes `quire new` pathing issues; `quire new` was previously installing `@thegetty/quire-11ty` in strange locations outside of `quire/packages/cli`
- Fixes broken paths in `symlinkLatest()` function
- Implements `getVersionFromStarterPeerDependencies()` to retrieve version of `@thegetty/quire-11ty` to install from the `peerDependencies` key of [`quire-starter-default`](https://github.com/thegetty/quire-starter-default) `package.json` (the version listed in that repository has been updated to `1.0.0-pre-release.2` to match the published version of `@thegetty/quire-11ty`)
- Implements `setVersion()` function to write the current version of `@thegetty/quire-11ty` being used with a new project to a `.quire` config file
- Adds notes and documentation of various `lib/quire` functions

### DIFF
--- a/packages/cli/src/lib/quire/index.js
+++ b/packages/cli/src/lib/quire/index.js
@@ -1,5 +1,5 @@
-import { cwd } from 'node:process'
-import { execa } from 'execa'
+import { chdir, cwd } from 'node:process'
+import { execa, execaCommand } from 'execa'
 import { fileURLToPath } from 'node:url'
 import { isEmpty } from '#helpers/is-empty.js'
 import fs from 'fs-extra'
@@ -91,15 +91,15 @@ async function initStarter (starter, projectPath) {
 
   // @TODO Remove 11ty file copying code (lines 79-88) once CLI pathing issues have been sorted out
   // Copy 11ty files
-  const fullProjectPath = path.resolve(projectPath)
-  const eleventyPath = path.resolve(path.join(INSTALL_PATH, quireVersion))
-  const eleventyFiles = fs.readdirSync(eleventyPath)
+  // const fullProjectPath = path.resolve(projectPath)
+  // const eleventyPath = path.resolve(path.join(INSTALL_PATH, quireVersion))
+  // const eleventyFiles = fs.readdirSync(eleventyPath)
 
   // copies all files in `quire/packages/11ty`
-  eleventyFiles.forEach((filePath) => {
-    const fileToCopy = path.resolve(eleventyPath, filePath)
-    fs.copySync(fileToCopy, path.join(fullProjectPath, path.basename(filePath)))
-  })
+  // eleventyFiles.forEach((filePath) => {
+  //   const fileToCopy = path.resolve(eleventyPath, filePath)
+  //   fs.copySync(fileToCopy, path.join(fullProjectPath, path.basename(filePath)))
+  // })
 
   // Reinitialize project as a new git repository
   await fs.remove(path.join(projectPath, '.git'))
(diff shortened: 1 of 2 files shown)


packages/cli/src/lib/quire/index.js
```diff
@@ -129,20 +129,44 @@ async function initStarter (starter, projectPath) {
  */
 async function install(version='latest', options={}) {
   console.debug(`[CLI:quire] installing quire-11ty@${version}`)
-  fs.ensureDirSync(INSTALL_PATH)
+  const absoluteInstallPath = path.join(__dirname, 'versions')
```
mphstudios, line 132: @anderspollack is this actually an _absolute_ path, would `path.isAbsolute(absoluteInstallPath)` be truthy? 
If this needs to be an absolute path we probably need something like `path.join(path.relative('/', __direname), 'version')`
which might be made into a utility module, perhaps 
```
import { isAbsolute, relative } from 'node:path'
export (path) => isAbsolute(path) ? path : relative('/', path)
```

packages/cli/src/lib/quire/index.js
```diff
@@ -129,20 +117,44 @@ async function initStarter (starter, projectPath) {
...
+  const invNodeModulesDir = path.join(absoluteInstallPath, 'node_modules')
+  if (fs.existsSync(invNodeModulesDir)) fs.rmdir(invNodeModulesDir)
+
   symlinkLatest()
+
+  console.debug('[CLI:quire] installing dev dependencies')
+  /**
+   * Install dev dependencies manually, as they are necessary to run 11ty.
+   * They should be kept as dev dependencies so that they don't get bundled into
+   * the final compiled `_site` package when running `quire build`
```
mphstudios, line 147:
```suggestion
   * Manually install necessary dev dependencies to run 11ty;
   * these must be `devDependencies` so that they are not bundled into
   * the final `_site` package when running `quire build`
```

packages/cli/src/lib/quire/index.js
```diff
@@ -127,28 +141,60 @@ async function initStarter (starter, projectPath) {
...
+   */
+  const currentWorkingDirectory = cwd()
+  const versionDir = path.join(absoluteInstallPath, version)
+  chdir(versionDir)
+  await execaCommand(
+    'npm cache clean --force'
+  ).stdout.pipe(process.stdout)
+  await execaCommand(
+    'npm install --save-dev'
+  ).stdout.pipe(process.stdout)
```
mphstudios, line 183: I do not think the standard out from these method calls will be displayed, we might remove it for now and revisit if we implement verbose output.
```suggestion
  await execaCommand('npm cache clean --force')
  await execaCommand('npm install --save-dev')
```

packages/cli/src/lib/quire/index.js
```diff
@@ -49,14 +49,34 @@ function getVersion(projectPath) {
...
+ *
+ * @return  {String}  version  Quire-11ty semantic version with caret or other
+ * comparators trimmed off the beginning
+ *
+ * @TODO refactor `latest()` function to programmatically return a specific
+ * version of `@thegetty/quire-11ty` from a semantic version string
+ * (i.e `^1.0.0-pre-release.0` => `1.0.0-pre-release.2`) so this string-trimming
+ * logic can be removed
+ */
+function getVersionFromStarterPeerDependencies(projectPath) {
```
mphstudios, line 65: Though this is very explicit it may be _too_ explicit, for example we might need to `getVersionFromStarter` but not do so from `peerDependencies`, perhaps removing the implementation detail from the method name,
```suggestion
function getVersionFromStarter(projectPath) {
```

mphstudios (approved):
@anderspollack this is nice work, see two inline comments on method names and `execa` output for your consideration.

PR #588, opened 2022-11-14
Fix `filterSymlinkLatest()`

anderspollack:
Not sure if this is a necessary extra step, since I have been monitoring the install process in `quire new` from my Finder (which will create `.DS_Store`s that normally wouldn't ever get made) but aside from `.DS_Store` being the thing returned from
``` js
  const latestInstalledVersion = fs
    .readdirSync(path.join(__dirname, 'versions'))
    .sort(semver.rcompare)[0] // <-- right here
```
`latest` was causing problems as not being a `semver.rcompare`-able thing

### DIFF
--- a/packages/cli/src/lib/quire/index.js
+++ b/packages/cli/src/lib/quire/index.js
@@ -249,7 +249,7 @@ function setVersion(projectPath, version) {
 function symlinkLatest() {
   const latestInstalledVersion = fs
     .readdirSync(path.join(__dirname, 'versions'))
-    .filter((filename) => !['.DS_Store', 'desktop.ini', 'thumbs.db', 'latest'].includes(filename))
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/quire/index.js
```diff
@@ -249,6 +249,7 @@ function setVersion(projectPath, version) {
 function symlinkLatest() {
   const latestInstalledVersion = fs
     .readdirSync(path.join(__dirname, 'versions'))
+    .filter((filename) => !['.DS_Store', 'desktop.ini', 'thumbs.db', 'latest'].includes(filename))
```
mphstudios, line 252:
```suggestion
    .filter((dirent) => semver.valid(semver.coerce(dirent))
```

PR #591, opened 2022-11-16
Quire CLI Plan B

anderspollack:
This PR implements our "Plan B" -- running quire projects with `@thegetty/quire-11ty` package decoupled from user content files has a few outstanding pathing issues, so this update allows users to create a new quire project where the `@thegetty/quire-11ty` files are added directly to the project and committed.

**Note:** the install process in `installInProject()`has to first install `@thegetty/quire-11ty` into a `.temp` directory in the `projectRoot`, then copy those files into the `projectRoot` as `install-npm-version` will not let you install things into a non-empty directory.

To use, run `quire new --eject [project name]` (open to suggestions on the name, as we are not really "ejecting" while creating)

This likely warrants an entirely new command `quire eject` which can be run on a previously created project.

- Implement and export `quire.installInProject` function in `lib/quire`
- Update JSDoc comment for `quire.install` function
- Remove any OS-specific slash characters from `quire.install`
- Add `--eject` option flag to `quire new` command
- Correctly display command name if `options.debug` is true
- Run `quire.installInProject` if `options.eject` (`--eject` flag) is passed to `quire new`

### DIFF
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -23,7 +23,7 @@ export default class CreateCommand extends Command {
     ],
     options: [
       [ '--debug', 'debug the `quire new` command' ],
-      [ '--eject', 'install quire-11ty into the project directory' ],
+      [ '--eject', 'install quire-11ty into the project directory', 'true' ],
     ],
   }
 
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -53,6 +53,9 @@ export default class CreateCommand extends Command {
       // `git clone starter path`
     } else {
       const version = await quire.initStarter(starter, projectPath)
+      // @TODO we will want to abstract the test for emptiness to prevent further install steps on error
+      if (!version) return
+
       if (options.eject) {
         await quire.installInProject(projectPath, version, options)
       }
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -23,6 +23,7 @@ export default class CreateCommand extends Command {
     ],
     options: [
       [ '--debug', 'debug the `quire new` command' ],
+      [ '--eject', 'create new quire project with 11ty files committed directly' ],
     ],
   }
 
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -58,8 +58,9 @@ export default class CreateCommand extends Command {
 
       if (options.eject) {
         await quire.installInProject(projectPath, version, options)
+      } else {
+        await quire.install(version, options)
       }
-      await quire.install(version, options)
     }
   }
 }
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -23,7 +23,7 @@ export default class CreateCommand extends Command {
     ],
     options: [
       [ '--debug', 'debug the `quire new` command' ],
-      [ '--eject', 'create new quire project with 11ty files committed directly' ],
+      [ '--eject', 'install quire-11ty into the project directory' ],
     ],
   }
 
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -53,7 +53,10 @@ export default class CreateCommand extends Command {
       // `git clone starter path`
     } else {
       const version = await quire.initStarter(starter, projectPath)
-      await quire.install(version)
+      if (options.eject) {
+        await quire.installInProject(projectPath, version, options)
+      }
+      await quire.install(version, options)
     }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/create.js
```diff
@@ -23,6 +23,7 @@ export default class CreateCommand extends Command {
     ],
     options: [
       [ '--debug', 'debug the `quire new` command' ],
+      [ '--eject', 'create new quire project with 11ty files committed directly' ],
```
mphstudios, line 26:
```suggestion
      [ '--eject', 'install quire-11ty into the project directory' ],
```

mphstudios (commented):
@anderspollack this looks good, plan-b for the win.

We might refactor the install methods to encapsulate the specific concerns, pseudo-code for refactoring `lib/quire`
```javascript
{
  initStarter (projectPath, starter, options = {}) {
    // install starter to projectPath ...
    const version = getVersionFromStarter()
    const installPath = options.eject ? 
      ? installInProject(version, projectPath, options)
      : install(version)
      
  }

  install (version, path, options = {}) {
    path ||= `./lib/versions/${version}`
    await inv = install(path)
    return path
  }

  installInProject (version, projectRoot, options = {}) {
    // path is the projectRoot
    const tempPath = install(version, path, options)
    // copy from tempPath to projectRoot
    return projectPath
  }
}
```

packages/cli/src/lib/quire/index.js
```diff
@@ -184,6 +183,74 @@ async function install(version, options={}) {
...
+  /**
+   * delete `package.json` from starter project, as it will be replaced with
+   * `package.json` from `@thegetty/quire-11ty`
+   */
+  await git
+    .cwd(projectPath)
+    .rm(['package.json'])
+    .catch((error) => console.error('[CLI:error] ', error))
+
+  const temp11tyDirectory = '.temp'
```
mphstudios, line 206: We might want to use a npm package that handles various edge cases around creating the temporary directory. For example if the install to the temp directory fails we need to ensure that we cleanup the file system and notify the user.

packages/cli/src/lib/quire/index.js
```diff
@@ -184,6 +183,74 @@ async function install(version, options={}) {
...
+   * @see https://github.com/scott-lin/install-npm-version
+   */
+  const installOptions = {
+    Destination: path.join('..', temp11tyDirectory),
+    Debug: false,
+    Overwrite: options.force || options.overwrite || false,
+    Verbosity: options.debug ? 'Debug' : 'Silent',
+    WorkingDirectory: projectPath
+  }
+  await inv.Install(`${PACKAGE_NAME}@${version}`, installOptions)
```
mphstudios, line 220: How does `inv` handle errors? We may want a `try` `catch` `finally` here.

packages/cli/src/lib/quire/index.js
```diff
@@ -184,6 +183,74 @@ async function install(version, options={}) {
...
+  fs.copySync(path.join(projectPath, '.temp'), projectPath)
+
+  console.debug('[CLI:quire] installing dev dependencies into quire project')
+  /**
+   * Manually install necessary dev dependencies to run 11ty;
+   * these must be `devDependencies` so that they are not bundled into
+   * the final `_site` package when running `quire build`
+   */
+  await execaCommand('npm cache clean --force', { cwd: projectPath })
+  await execaCommand('npm install --save-dev', { cwd: projectPath })
```
mphstudios, line 236: Let's use `try` the `npm` command, `catch` `error` and report to user, and `finally` cleanup the file system.

packages/cli/src/lib/quire/index.js
```diff
@@ -184,6 +183,74 @@ async function install(version, options={}) {
...
+    Verbosity: options.debug ? 'Debug' : 'Silent',
+    WorkingDirectory: projectPath
+  }
+  await inv.Install(`${PACKAGE_NAME}@${version}`, installOptions)
+
+  // delete empty `node_modules` directory that `install-npm-version` creates
+  const invNodeModulesDir = path.join(projectPath, 'node_modules')
+  if (fs.existsSync(invNodeModulesDir)) fs.rmdir(invNodeModulesDir)
+
+  // Copy all files installed in `.temp` to projectPath
```
mphstudios, line 226: Init starter also initializes a new package config (`package.json`), will this overwrite that file?

packages/cli/src/lib/quire/index.js
```diff
@@ -184,6 +183,80 @@ async function install(version, options={}) {
...
+async function installInProject(projectPath, version, options={}) {
+  console.debug(`[CLI:quire] installing quire-11ty@${version} into ${projectPath}`)
+
+  /**
+   * delete `package.json` from starter project, as it will be replaced with
+   * `package.json` from `@thegetty/quire-11ty`
+   */
+  await git
+    .cwd(projectPath)
+    .rm(['package.json'])
```
mphstudios, line 203: If a user runs `quire eject` at a latter date we may want to merge the two package configs... for a later pull-request.

mphstudios (approved):

PR #593, opened 2022-11-17
Refactor `getVersionFromStarter` function to parse `latest` version of `@thegetty/quire-cli` peer dependency

anderspollack:
If `quire-starter-default` is set to use `peerDependencies: { '@thegetty/quire-11ty': 'latest' }`, the `getVersionFromStarter()` function in `lib/quire` erroneously returns `t` as a version (`'latest'.substr('latest'.search(/\d/)) => 't'`). This update calls now calls `latest()` from `getVersionFromStarter()` if the version returned from starter `package.json` is `latest`

**NOTE!!!** This is still hacky -- using `npm view @thegetty/quire-11ty version` does indeed return a single semver version string, but since this is intended for human view, there's no future guarantee that `npm` won't reformat the text readout of `npm view [package] version`. We still need to find a programmatic way to parse versions from packages.

### DIFF
--- a/packages/cli/src/lib/quire/index.js
+++ b/packages/cli/src/lib/quire/index.js
@@ -75,6 +75,7 @@ async function getVersionFromStarter(projectPath) {
  * Clone or copy a Quire starter project
  *
  * @param    {String}   starter   A repository URL or an absolute path to local starter
+ * @TODO resolve both absolute and relative paths to local starter repositories 
  * @param    {String}   projectPath  Absolute system path to the project root
  *
  * @return   {String}   quireVersion  A string indicating the current version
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/quire/index.js
```diff
@@ -62,17 +62,19 @@ function getVersion(projectPath) {
...
+  return version === 'latest'
+    ? await latest()
+    : version.substr(version.search(/\d/))
 }
 
 /**
  * Clone or copy a Quire starter project
  *
- * @param    {String}   starter   A repository URL or path to local starter
+ * @param    {String}   starter   A repository URL or an absolute path to local starter
```
anderspollack, line 77: **TODO**: resolve both absolute and relative paths to local starter directories in `initStarter`. I updated the JSDoc here since it wasn't clear why this wasn't working at first
```suggestion
 * @param    {String}   starter   A repository URL or an absolute path to local starter
 * @TODO resolve both absolute and relative paths to local starter repositories 
```

PR #595, opened 2022-11-21
Set Eleventy dir path environment variables

mphstudios:
Eleventy does not expose an API to set the `includes` and `layouts` paths on the `eleventy` instance. In order to separate Quire's Eleventy code from the Quire project content these paths need to be set in our `.eleventy.js` configuration file **before** creating the `eleventy` instance.

Currently this pull-request is blocked by the current implementation of the Eleventy `TemplatePathResolver`, see [@11ty/eleventy#2655](https://github.com/11ty/eleventy/issues/2655), which assumes that the `layouts` directory is a child of the `input` directory.

## Changed

- Sets process environment variables for Eleventy paths that are read by the eleventy configuration file.
- Use the Eleventy programmatic API by default when running the `quire` `build` and `preview` commands; running these commands using the Eleventy `cli` hangs the process.

## TODO

- [x] Fix process from hanging when `quire` `build` and `preview` commands are run using the Eleventy `cli`.


mphstudios:
@thegetty/int-dev though https://github.com/11ty/eleventy/issues/2655 prevents us making use of the path env vars at present this pull-request is ready; I would welcome your reviews and thoughts on wether we should merge these changes now?

hbalenda:
I'm cool with merging this now, maybe add a note with a link to the open Eleventy issue to the `packages/cli/src/lib/11ty/README.md` or top of `packages/cli/src/lib/11ty/cli.js`?

mphstudios:
> add a note with a link to the open Eleventy issue

Good call, added to the README and to each lib/11ty module.

PR #598, opened 2022-11-28
Delete `content` from `packages/11ty`

anderspollack:
Content for new projects is now provided by `quire-starter-default`; keeping this content in here was overwriting starter content when running `quire new [--eject=true]`

this will complicate old way of dev experience (`cd packages/11ty && npm run dev`) but I don't think that matters? can we publish `@thegetty/quire-11ty`with everything  _except_ this dummy content directory?


hbalenda (approved):

mphstudios (approved):
It is time, packages/11ty/~content~; we need to document the local development workflow.

PR #599, opened 2022-11-28
Update pagedjs-cli options for 0.3.0

fchasen:
Update the default CLI options for Pagedjs CLI.


packages/cli/src/lib/pdf/paged.js
```diff
@@ -9,15 +9,20 @@ export default async (input, options) => {
    * Configure the Paged.js Printer options
    * @see https://gitlab.coko.foundation/pagedjs/pagedjs-cli/-/blob/main/src/cli.js
    */
-  const printerOptions = { debug: options.debug || false }
+  const printerOptions = {
+    debug: options.debug || false,
+    allowLocal: true,
```
mphstudios, line 14:
```suggestion
    allowLocal: true,
    debug: options.debug || false,
```

mphstudios (approved):

PR #600, opened 2022-11-28
Fix/DEV-13676: Set eleventy env in `build` and `serve` commands

hbalenda:
Internally, `quire-11ty` uses the `ELEVENTY_ENV` environment variable with a value of `production` or `development` for conditional handling of writing files (for example: pdf files, search index) directly to the output directory in the `eleventy.after` hook so that they are not removed by vite. Previously `ELEVENTY_ENV` was set using `cross-env` in the npm `dev` and `build` scripts, and now needs to be included in the `execa` environment.

### DIFF
--- a/packages/cli/src/lib/11ty/api.js
+++ b/packages/cli/src/lib/11ty/api.js
@@ -94,6 +94,7 @@ export default {
     console.info('[CLI:11ty] running eleventy build')
     console.info(`[CLI:11ty] projectRoot ${projectRoot}`)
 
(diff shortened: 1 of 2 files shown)


mphstudios (changes requested):
@hbalenda though we are not using the `lib/api` by default, setting the `process.env.ELEVENTY_ENV` in the `lib/api.build` and `lib/api.serve` methods seems necessary as well for the `api` to run correctly, no?

mphstudios (approved):

PR #607, opened 2022-11-30
Test current working directory is a project

mphstudios:
Outputs a console message when a `quire` command must be run from the project directory.

## Added

- CLI helper to test the current working directory is a project root or exit the process with a failure code `1`.

## Changed

- CLI commands `build`, `clean`, `preview`, `version` run `testcwd` in a `preAction`

### DIFF
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -1,6 +1,7 @@
 import Command from '#src/Command.js'
 import { api, cli, paths, projectRoot  } from '#lib/11ty/index.js'
 import { clean } from '#helpers/clean.js'
+import testcwd from '#helpers/test-cwd.js'
 
 /**
  * Quire CLI `build` Command
@@ -56,6 +57,8 @@ export default class BuildCommand extends Command {
   }
 
   preAction(command) {
+    testcwd(command)
+
     const options = command.opts()
     if (options.debug) {
       console.debug('[CLI] Calling \'build\' command pre-action with options', options)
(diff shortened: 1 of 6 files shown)


geealbers:
@Erin-Cecele, this PR from @mphstudios will output a message if someone tries to run a Quire command but are not in a Quire project.

```
[CLI] preview command must be run from a Quire project directory
```
It's a good idea. But what do you think about the wording and information provided? Should we offer more? Not sure we can succinctly explain navigating a command-line shell, but maybe a mention of `cd`, or is there anywhere in the docs where we talk about this and should we include a link to that in this message? I'm not sure myself

mphstudios:
> It's a good idea. But what do you think about the wording and information provided? Should we offer more? ... or is there anywhere in the docs where we talk about this and should we include a link to that in this message?

I think that the wording could be improved and that including a link to the Quire docs would be great. It may be a good idea to do this holistically as a separate chunk of work post-pre-release.

Erin-Cecele:
@geealbers @mphstudios I love the idea of having that message and agree that it would be best to include instructions for how to use `cd` to get into the project directory. For now, maybe we could say:

```
[CLI] preview command must be run while in a Quire project directory. Use `cd` to navigate to your project directory and re-run the preview. 
``` 
Or something to that effect. And we can add "See the docs for more information" post-pre-release, as Matthew suggests. 

I have to think about where in the docs we would link to. Right now, `cd`, `ls`, etc. is most explicitly explained in the Tutorial. But I think we could add something to the Quire Commands section. "Useful Non-Quire Commands" or something like that. In any other section of the docs, this info will get buried.

PR #609, opened 2022-11-30
Format logging output for increased legibility

geealbers:
Adds `[quire]` prefix, log types, and column spacing. See https://jira.getty.edu/browse/DEV-13749.

### DIFF
--- a/packages/11ty/_lib/chalk/index.js
+++ b/packages/11ty/_lib/chalk/index.js
@@ -41,7 +41,8 @@ module.exports = function (prefix='', loglevel=2) {
   const logFn = (type) => {
     const log = logger[type]
     const style = styles[type]
-    return (message) => log(style(chalk.bold(`[${prefix}]`), message))
+    const prefixSpacer = prefix.length > 30 ? " " : " ".repeat(30 - prefix.length)
+    return (message) => log(style(chalk.bold(`[quire] ${type.toUpperCase()}\t`), `${prefix}${prefixSpacer}`, message))
   }
 
   /**
--- a/packages/11ty/_lib/chalk/index.js
+++ b/packages/11ty/_lib/chalk/index.js
@@ -41,8 +41,8 @@ module.exports = function (prefix='', loglevel=2) {
   const logFn = (type) => {
     const log = logger[type]
     const style = styles[type]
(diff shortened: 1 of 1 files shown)


packages/11ty/_lib/chalk/index.js
```diff
@@ -41,7 +41,8 @@ module.exports = function (prefix='', loglevel=2) {
   const logFn = (type) => {
     const log = logger[type]
     const style = styles[type]
-    return (message) => log(style(chalk.bold(`[${prefix}]`), message))
+    const prefixSpacer = prefix.length > 30 ? " " : " ".repeat(30 - prefix.length)
+    return (message) => log(style(chalk.bold(`[quire] ${type.toUpperCase()}\t`), `${prefix}${prefixSpacer}`, message))
```
mphstudios, line 45: @geealbers it may work nicely to use the `String.prototype.padEnd()` method, which accepts a target length and the padding character. 
```suggestion
    prefix = prefix.padEnd(30, '\u0020')
    return (message) => log(style(chalk.bold(`[quire] ${type.toUpperCase()}\t`), `${prefix}`, message))
```

geealbers: Ah yes, that's nicer, and works in testing. Thanks!

geealbers:
@mphstudios You asked about what this looks like when the line breaks and you're right it is wonky:

Screen Shot 2022-11-30 at 2 34 03 PM

Though it's pretty wonky even without the columns. Down the road we may want to look at something like https://www.npmjs.com/package/indent-log for better indenting with line breaks:

log3

mphstudios:
> Down the road we may want to look at something like [npmjs.com/package/indent-log](https://www.npmjs.com/package/indent-log) for better indenting with line breaks

Absolutely, something like `indent-log` and/or `cliui` to better format the output. I also think that we will be able to bring back the fancy colours.

@geealbers @Erin-Cecele @thegetty/int-dev what are your thoughts on including the changes in this pull-request in the pre-release and refactoring later?

geealbers:
@mphstudios, it makes sense to me to merge this in for the pre-release as is, and then refactor it later

Erin-Cecele:
@mphstudios @geealbers @thegetty/int-dev Despite it being hard to read with the line breaks, I think overall, the messaging is clearer and more helpful. I'd love to see this merged for the pre-release and refactored later.

PR #623, opened 2022-12-08
Refactor/DEV-13794: EPUB nav element

hbalenda:
Change:
- Changes TOC wrapper tag from `div` to `nav` (semantic!)
- Adds special EPUB attributes to table of contents element during EPUB transform


mphstudios (approved):
Semantic BEM style for the win.

PR #631, opened 2022-12-15
Fix duplicate id attributes for EPUB and PDF

mphstudios:
Ensure that the transformed output for EPUB and PDF generation does not have heading or section elements with duplicate `id` attributes.

### Fixed

- Uses `slugify` on the page title when setting a value for `h1` `id` attributes.
- Corrects misspelling in JSDoc comments.


packages/11ty/_plugins/transforms/outputs/pdf/write.js
```diff
@@ -69,5 +72,28 @@ module.exports = (eleventyConfig) => {
...
+      return duplicates
+    }, [])
+
+    duplicateIds.forEach((id) => {
+      const elementsWithSameId = Array.from(document.querySelectorAll(`[id=${id}]`))
+      for (const [index, element] of elementsWithSameId.entries()) {
+        if (index) {
+          element.id += `_${index}`
+        }
+      }
```
mphstudios, line 96:
```
      elementsWithSameId.forEach(([element, index]) => {
        if (index) element.id += `_${index}`
      })
```
though not certain why would not have an `index`, so perhaps
```suggestion
      elementsWithSameId.forEach(([element, index]) => element.id += `_${index}`)
```

packages/11ty/_layouts/entry.liquid
```diff
@@ -47,7 +47,7 @@ Entry content, including entry image and tombstone data
   <div class="quire-entry__content">
     <header class="quire-entry__header">
       <div class="container">
-        <h1 class="quire-page__header__title" id="{{ title | url }}">
+        <h1 class="quire-page__header__title" id="{{ title | slugify }}">
```
mphstudios, line 50: @anderspollack @fchasen we might use `page.filePathStem` or `page.fileSlug` here to to prevent duplicates though remove the file extensions, perhaps implementing an `idSlug` filter, somthing like

```javascript
  const { dir, name } path.parse(page.fileSlug)
  return `${dir}/${name}`
```

```suggestion
        <h1 class="quire-page__header__title" id="{{ page.fileSlug | idSlug }}">
```

fchasen: This make sense to me, but we might want to append some text like "header_" to the file path in case it starts with a number.

packages/11ty/_plugins/transforms/outputs/pdf/write.js
anderspollack: yeah this is a little funny, the `if (index)` is checking that the index is non-zero, not for the presence of an index -- this is so only subsequent duplicates get the `_#` digit appended. Maybe being more explicit with `if (index > 0)` explains intent better?
```suggestion
      elementsWithSameId.forEach((element, index) => {
        if (index > 0) element.id += `_${index}`
      })
```

anderspollack:
@mphstudios I moved the ID de-duplication logic into a standalone `dedupeElementIds.js` module used by `outputs/pdf/write.js` and `outputs/epub/transform.js`

packages/11ty/_plugins/transforms/outputs/pdf/write.js
```diff
@@ -57,7 +63,7 @@ module.exports = (eleventyConfig) => {
     const sassOptions = {
       loadPaths: [
         path.resolve('node_modules')
-      ]
+      ],
```
anderspollack, line 66:
```suggestion
      ]
```

packages/11ty/_plugins/transforms/outputs/pdf/write.js
mphstudios: the version with the explicit `index > 0` is better, though with JSDoc would be even better still.

packages/11ty/_plugins/markdown/index.js
```diff
@@ -22,8 +22,11 @@ const removeMarkdown = require('remove-markdown')
 module.exports = function(eleventyConfig, options) {
   /**
    * @see https://github.com/valeriangalliat/markdown-it-anchor#usage
+   * To prevent duplicate element IDs from slugified headings, we are only generating anchor links for level 1 headings
    */
-  const anchorOptions = {}
+  const anchorOptions = {
+    level: [1]
+  }
```
anderspollack, line 29: @mphstudios this is a kinda dumb hack to remove duplicate IDs -- configuring automatic anchor linking to only occur on H1s -- might be better to just remove auto-anchor linking if that's what we really want to happen?

It would be nice to set this configuration per-output, but as far as I can tell that is not possible given the current way we handle different outputs individually in components. We could also strip out all H2 ids across the board in PDF output in the transform as well?

mphstudios: @anderspollack I think this is good solution. Let's merge these changes.

anderspollack (approved):

PR #639, opened 2022-12-23
Handle null values for src

hbalenda:
Prevents preview from failing due to `null` value for `figure.src`


packages/11ty/_includes/components/figure/image/print.js
```diff
@@ -1,5 +1,7 @@
 const { html } = require('~lib/common-tags')
 const path = require('path')
+const chalkFactory = require('~lib/chalk')
+const logger = chalkFactory('Figure Annotations UI')
```
hbalenda, line 4:
```suggestion
```

anderspollack (approved):

PR #645, opened 2023-01-10
DEV-13196: Update cover and splash page images to render with relative image urls in PDF output

anderspollack:


packages/11ty/_plugins/transforms/outputs/pdf/write.js
```diff
@@ -48,6 +48,11 @@ module.exports = (eleventyConfig) => {
       asset.setAttribute('src', trimLeadingSlash(src))
     })
 
+    document.querySelectorAll('[data-background-image]').forEach((element) => {
+      const { backgroundImage } = element.dataset
+      if (backgroundImage) element.style.backgroundImage = `url('${trimLeadingSlash(backgroundImage)}')`
+    })
```
mphstudios, line 54: @anderspollack were we to execute this transform for all `style` background images we would not need to duplicate `data-background-image` attribute;
```suggestion
    document.querySelectorAll('[style*='background-image']').forEach((element) => {
      const { backgroundImage } = element.style
      if (backgroundImage) element.style.backgroundImage = `url('${trimLeadingSlash(backgroundImage)}')`
    })
```

mphstudios (approved):

PR #661, opened 2023-02-10
Update to Eleventy 2.0.0

mphstudios:
## Bumps

- Update to `eleventy@2.0.0` stable
- Update to `eleventy-img@3.0.0`
- Update `@iiif/parser` patch version
- Update `@iiif/vault` patch version


anderspollack (approved):
not seeing any issues with updated deps in a test pub, looks good to me

PR #663, opened 2023-02-14
JSON schemas for Quire data

mphstudios:
**_Work in progress_** to implement schemas for Quire data

### Added

- [Ajv](https://ajv.js.org) to `devDependencies` to validate data against schemas
- _draft_ schema for `figures` data

Quire data can be validated against the schema using the `ajv-cli`, for example
```sh
npx ajv validate -s packages/11ty/_plugins/figures/schema.json -d ~/Code/Getty/quire-starter-default/content/_data/figures.yaml
```
however an error thrown when the `$schema` key is included,
```sh
error: no schema with key or ref "https://json-schema.org/draft/2020-12/schema"
```
a temporary work around is to remove the `$schema` property on line 3 of `figures/schema.json`


packages/11ty/_plugins/figures/schema.json
```diff
@@ -0,0 +1,175 @@
...
+  "type": "object",
+  "definitions": {
+    "Figures": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "figure_list": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/FigureList"
```
hbalenda, line 16:
```suggestion
            "$ref": "#/definitions/Figure"
```
Might be more clear if this object definition is `Figure` since it's the item not the list?

packages/11ty/_plugins/figures/schema.json
```diff
@@ -0,0 +1,175 @@
...
+        }
+      },
+      "required": [
+        "figure_list"
+      ],
+      "title": "Figures"
+    },
+    "FigureList": {
+      "type": "object",
+      "additionalProperties": false,
```
hbalenda, line 27: I wonder if we should allow additional properties - this might be a key area where people who are using Quire in an off-menu way might expand on the project. I guess they could go in and edit this schema if they wanted to do that... but just a thought.

mphstudios: @hbalenda allowing additional properties is a good idea; initially I think we can set it to `false` and then either update the schema to allow them or perhaps implement this as a Quire project setting that treats additional properties as warnings rather than errors when we are validating the figures data.

PR #665, opened 2023-02-15
DEV-14144: Process image sequence files

anderspollack:
This implements a `processFigureSequence` instance method for the `Figure` class to generate tiles for image sequences. Additionally, there are now 2 more helpers:
- `isSequence` determines if a figure has a `sequence` property
- `getSequenceFiles` looks in the directory matching the `sequence` id to fetch all sequence image file names

**Note:** The branching logic introduced here feels a little hacky, and the `Figure` should probably get split into sub-classes once all of the manifest-generation logic for image sequences is working, but I wanted to open a PR for this so all this 360 stuff doesn't end up in a massive mega-PR.


hbalenda (approved):
Looks great!

PR #666, opened 2023-02-15
DEV-14145: Image sequence manifests

anderspollack:
**WIP** Refactoring IIIF manifest module to write image sequence manifests


packages/11ty/_plugins/figures/figure/index.js
```diff
@@ -74,9 +76,11 @@ module.exports = class Figure {
     this.outputDir = outputDir
     this.processImage = imageProcessor
     if (this.isSequence) {
-      this.sequenceDir = data.sequence[0].id
+      this.sequenceDir = data.sequences[0].id
+      this.sequenceFactory = new SequenceFactory(this)
       this.sequenceFiles = getSequenceFiles(data, iiifConfig)
-      this.sequenceRegex = data.sequence[0].regex
+      this.sequenceRegex = data.sequences[0].regex
```
hbalenda, line 82: If we're supporting multiple sequences, don't we want to iterate over `data.sequences` and run all the sequence stuff on each one, similar to how we handle annotations? Or are we just supporting the first sequence?

PR #668, opened 2023-02-17
Refactor GlobalData and Vite plugins

hbalenda:
Changes:
- Encapsulate `EleventyPluginVite` config in `~plugins/vite`
- Return global data from `globalData` plugin. To do this, the plugin has to be immediately invoked, which requires that `dataExtensions` must be called before `globalData`.
- Add some (minor) validation to `publication.yaml` that ensures the `url` is valid and adds a `pathname` property.
- Run Figures plugin with the rest of the plugins instead of during `eleventy.before` hook so that it has access to vite config

### DIFF
--- a/packages/11ty/.eleventy.js
+++ b/packages/11ty/.eleventy.js
@@ -120,6 +120,10 @@ module.exports = function(eleventyConfig) {
    *   1) immediately invoked
    *   2) addPlugin statements
    * Plugins that mutate globalData must be added before other plugins
+   *
+   * Note: The config does **not** have access to collections or global,
+   * to get around this we invoke some plugins immediately and return a value
+   * so that data can be provided to the config or another plugin.
    */
   dataExtensionsPlugin(eleventyConfig)
   const globalData = globalDataPlugin(eleventyConfig, { inputDir })
--- a/packages/11ty/.eleventy.js
+++ b/packages/11ty/.eleventy.js
@@ -5,7 +5,6 @@ const fs = require('fs-extra')
 const packageJSON = require('./package.json');
 const path = require('path')
 const scss = require('rollup-plugin-scss')
-const yaml = require('js-yaml');
 
 const chalkFactory = require('~lib/chalk')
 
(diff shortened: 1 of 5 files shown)


packages/11ty/_plugins/vite/index.js
```diff
@@ -0,0 +1,90 @@
...
+module.exports = function(eleventyConfig, { inputDir, outputDir, globalData }) {
+  /**
+   * Use Vite to bundle JavaScript
+   * @see https://github.com/11ty/eleventy-plugin-vite
+   *
+   * Runs Vite as Middleware in the Eleventy Dev Server
+   * Runs Vite build to postprocess the Eleventy build output
+   */
+  const { pathname } = globalData.publication
+  const pathResolutionAliases = pathname === '/' ? [] : [{ find: pathname, replacement: '/' }]
```
mphstudios, line 14: Moving this expression to line 28 to set the `alias` property directly may be more explicit.

packages/11ty/.eleventy.js
```diff
@@ -151,11 +116,16 @@ module.exports = function(eleventyConfig) {
   eleventyConfig.addPlugin(EleventyHtmlBasePlugin)
 
   /**
-   * Plugins are loaded in order of the `addPlugin` statements,
-   * plugins that mutate globalData must be added before other plugins
+   * Plugins are loaded in this order: 
+   *   1) immediately invoked
+   *   2) addPlugin statements
+   * Plugins that mutate globalData must be added before other plugins
```
mphstudios, line 122: Expanding on this comment with more detail about how each of these plugins rely on data from the previous may be kind to our future selves.

hbalenda: Yeah, it is confusing. I've updated the comment

packages/11ty/.eleventy.js
```diff
@@ -151,11 +116,20 @@ module.exports = function(eleventyConfig) {
...
+   *
+   * Note: The config does **not** have access to collections or global,
+   * to get around this we invoke some plugins immediately and return a value
+   * so that data can be provided to the config or another plugin.
```
mphstudios, line 126: Comments for the win; this makes it not-confusing.

mphstudios (approved):

PR #681, opened 2023-03-06
Add a version option for the quire-cli new command

mphstudios:
### Changes

Refactor the `quire new` command to have a `--quire <version>` option for the `quire-11ty` version that will be installed when creating a project; the default value is the npm distribution tag `latest`.

**Nota bene:** this is a _work-in-progess_ implementation that is currently intended for development only to test release candidate versions; this implementation does **not** check that the starter project is compatible with the `--quire <version>` option, see `@TODO` comments.

### DIFF
--- a/packages/cli/src/lib/quire/index.js
+++ b/packages/cli/src/lib/quire/index.js
@@ -148,7 +148,7 @@ async function initStarter (starter, projectPath) {
  * @return  {Promise}
  */
 async function install(options = {}) {
-  const version = options.version || 'latest'
+  const version = options.quire || 'latest'
   console.debug(`[CLI:quire] installing quire-11ty@${version}`)
   const absoluteInstallPath = path.join(__dirname, 'versions')
   fs.ensureDirSync(absoluteInstallPath)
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/quire/index.js
```diff
@@ -185,19 +187,22 @@ async function install(version, options={}) {
...
+ * @TODO refactor this to use the install method with pre and post-install hooks
+ * or steps to prepare the working directory and cleanup on error and completion
  *
  * @param  {String}  projectPath  Absolute system path to the project root
- * @param  {String}  version  Quire-11ty semantic version
  * @param  {Object}  options  options passed from `quire new` command
  * @return  {Promise}
  */
-async function installInProject(projectPath, version, options={}) {
+async function installInProject(projectPath, options = {}) {
```
anderspollack, line 199: the call to `setVersion` in `installInProject` is setting the `.quire` version to still use latest, rather than passing the specified version from `--quire` flag.

Output when I generate a new pub:
```sh
[CLI:quire] init-starter 
  project root: "blerg" 
  starter: "https://github.com/thegetty/quire-starter-default"
blerg set to use quire-11ty@1.0.0-rc.3  # THIS IS STILL 1.0.0-rc.3
[CLI:quire] installing quire-11ty@1.0.0-rc.5 into blerg
[CLI:quire] installing dev dependencies into quire project
```

anderspollack (approved):

PR #682, opened 2023-03-06
Manifest tests and data validation

anderspollack:
- Adds fixture data for figures, iiifConfig, and a manifest


packages/11ty/_plugins/figures/tests/__fixtures__/figures/zoom/manifest.json
```diff
@@ -0,0 +1,58 @@
+{
+  "@context": "http://iiif.io/api/presentation/3/context.json",
+  "id": "http://localhost:8080/iiif/zoom/manifest.json",
+  "type": "Manifest",
+  "behavior": [
+    [
+      "continuous",
+      "sequence"
+    ]
+  ],
```
mphstudios, line 10: The `behavior` property is only valid on `sequences`, no?

anderspollack: true!

packages/11ty/_plugins/figures/tests/__fixtures__/figures/zoomable/manifest.json
```diff
@@ -0,0 +1,58 @@
+{
+  "@context": "http://iiif.io/api/presentation/3/context.json",
+  "id": "http://localhost:8080/iiif/zoom/manifest.json",
+  "type": "Manifest",
+  "behavior": [
+    [
+      "continuous",
+      "sequence"
+    ]
+  ],
```
mphstudios, line 10: @anderspollack unless I am mistaken the `sequence` value for [behavior](https://iiif.io/api/presentation/3.0/#behavior) is only 
> valid only on Ranges, where the Range is [referenced](https://iiif.io/api/presentation/3.0/#12-terminology) in the structures property of a Manifest.
Interestingly, the following part of the the spec provides guidance on how a figure with multiple sequences might indicate the UI to _page_ through the sequences that are each `continous` --
> A Range with the behavior value continuous, in a Manifest with the behavior value paged, would mean that the Manifest’s Canvases should be rendered in a paged fashion, unless the range is selected to be viewed, and its included Canvases would be rendered in that context only as being virtually stitched together.

mphstudios (commented):
@anderspollack unless I am mistaken the `sequence` value for [behavior](https://iiif.io/api/presentation/3.0/#behavior) is only 
> valid only on Ranges, where the Range is [referenced](https://iiif.io/api/presentation/3.0/#12-terminology) in the structures property of a Manifest.

In all of the manifests, the `"behavior"` property should be set on the Range rather than as a top-level property. The exception might be when a figure has  multiple `sequence` structures and we were to want to indicate that the UI should _page_ through these sequences, in this case the manifest could have a `"behavior": ["paged"]` property and each `sequence` a `"behavior": ["continuous", "sequence"]` property; from the spec:

>A Range with the behavior value continuous, in a Manifest with the behavior value paged, would mean that the Manifest’s Canvases should be rendered in a paged fashion, unless the range is selected to be viewed, and its included Canvases would be rendered in that context only as being virtually stitched together.

packages/11ty/_plugins/figures/tests/index.js
```diff
@@ -0,0 +1,51 @@
...
+  const figure = new Figure(iiifConfig, null, figureFixture)
+  figure.canvasHeight = height
+  figure.canvasWidth = width
+  const manifest = new Manifest(figure)
+  const manifestJson = await manifest.toJSON()
+  return {
+    manifestFixture,
+    manifestJson
+  }
+}
```
mphstudios, line 25: @anderspollack this is an interesting approach for the unit testing the input/output; it seems to isolate that the for the given `figure.json` **input** data the `Figure` class will return an instance from which the `Manifest` will **output** the correct `manifest.json`. 

The integration test likely need to **mock** the file system, either using the [`mock-fs`](https://www.npmjs.com/package/mock-fs) package and for integration tests with a starter template using the [`unionfs`](https://github.com/streamich/unionfs) and [`memfs`](https://github.com/streamich/memfs) packages to merge the actual file system with the in-memory mocks.

packages/11ty/package.json
```diff
@@ -32,9 +32,10 @@
     "lint:fix": "npm run lint -- --fix",
     "prebuild": "npm run clean",
     "predev": "del-cli public",
+    "prepublishOnly": "npm run test",
     "preserve": "npm run clean",
     "serve": "cross-env ELEVENTY_ENV=production eleventy --serve",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "node --test _plugins/figures/tests/index.js"
```
hbalenda, line 38:
```suggestion
    "test": "node --test"
```
By defualt `node --test` will run everything in a `test` directory, so if you rename the `11ty/_plugins/figures/tests` directory to `11ty/_plugins/figures/test` and we continue to follow that pattern, this command will run the (future) whole 11ty test suite

anderspollack: ah nice!

PR #690, opened 2023-03-09
Add directoryConfig to globalData

hbalenda:
The directory config is not available to immediately invoked plugins (like vite), so previously we were passing directories explicitly. Additionally, the `publicDir` was not available in the global data at all, so the `iiifConfig` was previously parsing the vite options to get it.

Changes:
- Provide `inputDir` `outputDir` and `publicDir` to globalData plugin and add them to globalData under `directoryConfig` key
- Decouple `iiifConfig` from vite plugin

### DIFF
--- a/packages/11ty/.eleventy.js
+++ b/packages/11ty/.eleventy.js
@@ -39,7 +39,7 @@ const { error } = chalkFactory('eleventy config')
 
 const inputDir = process.env.ELEVENTY_INPUT || 'content'
 const outputDir = process.env.ELEVENTY_OUTPUT || '_site'
-const publicDir = 'public'
+const publicDir = process.env.ELEVENTY_ENV === 'production' ? 'public' : false // yes, false
 
 /**
  * Eleventy configuration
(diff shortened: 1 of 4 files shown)


packages/11ty/.eleventy.js
```diff
@@ -39,7 +39,7 @@ const { error } = chalkFactory('eleventy config')
 
 const inputDir = process.env.ELEVENTY_INPUT || 'content'
 const outputDir = process.env.ELEVENTY_OUTPUT || '_site'
-const publicDir = 'public'
+const publicDir = process.env.ELEVENTY_ENV === 'production' ? 'public' : false // yes, false
```
mphstudios, line 42: What does `// yes, false` mean?

mphstudios (commented):
@hbalenda it looks like `eleventyConfig.eleventyFiles` is available to the plugins though does not have a `publicDir` property?

packages/11ty/.eleventy.js
hbalenda, line 42:
```suggestion
const publicDir = process.env.ELEVENTY_ENV === 'production' ? 'public' : false // publicDir should be set explicitly to false in development 
```

mphstudios (approved):

hbalenda:
@mphstudios that property is not available to the plugins that we invoke without using `addPlugin` (for example, the vite plugin)

PR #695, opened 2023-03-16
DEV 14587: Custom WebC `<image-sequence>` implementation

anderspollack:
- Don't display sequence panels inside a `figureModalLink`
- Pre-load non-zoom sequence images
- Update @11ty/eleventy-plugin-webc
- Disable zoom for sequences
- Join annotaion ui items to prevent ',' in markup
- Add support for async components to `render`; join markup with newline
- Add async support to figure image
- Add custom `<image-sequence>` webc component, `figureImageSequence` quire component
- Implement `renderWebcComponent` filter to work around `eleventy-plugin-webc` not working when we use custom elements in shortcodes


packages/11ty/.eleventy.js
```diff
@@ -170,11 +170,11 @@ module.exports = function(eleventyConfig) {
...
-  eleventyConfig.addPlugin(pluginWebc, {
-    components: '_includes/components/**/*.webc',
-    transformData: {},
-    useTransform: false,
-  })
+  // eleventyConfig.addPlugin(pluginWebc, {
+  //   components: '_includes/components/**/*.webc',
+  //   transformData: {},
+  //   useTransform: false,
+  // })
```
anderspollack, line 177: This is not using `pluginWebc` at the moment, `figure/image/sequence.js` uses the webc API directly

packages/11ty/_includes/components/figure/annotations-ui/index.js
```diff
@@ -21,15 +21,16 @@ module.exports = function(eleventyConfig) {
...
+        .map((annotation, index) => figureOption({ annotation, index, input, name }))
+        .join('\n')
       return html`
         <fieldset class="annotations-ui">
           <legend>${title}</legend>
           ${options}
         </fieldset>
       `
-    })
+    }).join('\n')
```
anderspollack, line 33: a few things were not being joined, potentially leading to stray `,` gremlins

packages/11ty/_includes/components/figure/image/sequence-panel.js
```diff
@@ -26,10 +25,10 @@ module.exports = function(eleventyConfig) {
       id,
       manifestId,
       margin='',
-      preset='responsive',
+      preset='static',
       startCanvas,
-      textEnabled='true',
-      textSelectionEnabled='true',
+      textEnabled='false',
+      textSelectionEnabled='false',
```
anderspollack, line 31: these don't actually do anything, left over from when I was trying to make digirati `<sequence-panel>` work

packages/11ty/_includes/components/head.js
```diff
@@ -24,6 +25,20 @@ module.exports = function(eleventyConfig) {
...
+          const { ext } = path.parse(uri)
+          if (ext === '.json') {
+            // TODO lookup the tile defaults - need paths from sharp output
+            return ''
+          } else {
+            return `<link rel="preload" as="image" href="${uri}">`
+          }
+        }).join('\n')
+      }).join('\n')
+
```
anderspollack, line 41: This could be replaced with some fanciness in `image-sequence` `connectedCallback`, but this works well enough:
```
this.images = this.items.map((src) => {
  const image = new Image();
  image.src = src;
  return  image;
});
```

packages/11ty/_plugins/filters/renderWebcComponent.js
```diff
@@ -0,0 +1,13 @@
...
+  const { WebC } = await import('@11ty/webc')
+  const page = new WebC()
+  const { name: componentName } = path.parse(filePath)
+  const attributes = Object.keys(attrs).reduce((acc, key) => {
+    return acc += `${key}="${attrs[key]}"`
+  }, '')
+  page.setContent(`<${componentName} ${attributes} webc:import="${filePath}"></${componentName}>`)
+  const { html, css, js, components } = await page.compile()
+  return html
+}
```
anderspollack, line 13: this is doing what `eleventy-plugin-webc` should be doing for us

packages/11ty/_includes/components/head.js
mphstudios: We might add a `src` argument to the `Image` constructor so that we can write
```
this.images = this.items.map((src) => new Image(src));
```

packages/11ty/_includes/components/figure/annotations-ui/index.js
mphstudios: odd, perhaps when we refactor at some point we can remove the need for any chained `join('\n') method calls?

mphstudios (changes requested):
@anderspollack for each of the things in the code that you have called out in the pull-request comments add a comment in the code, either `@todo` or an explanation of what may not be easily understood about the implementation and the issues it addresses or works around.

packages/11ty/_includes/components/figure/image/image-sequence.webc
```diff
@@ -0,0 +1,221 @@
...
+        this.showImage(this.images[newIndex]);
+      }
+      const oldIndex = parseInt(oldValue);
+      if (oldIndex !== newIndex && !isNaN(oldIndex)) {
+        this.hideImage(this.images[oldIndex]);
+      }
+    }
+
+  });
+}
```
anderspollack, line 220: WebC is very strict about semicolons/no `//` comments/other things! It would be good developer experience to lint these files, since our build process won't catch client-side errors.

PR #697, opened 2023-03-17
Remove `urlFilter` from TOC list/grid item

anderspollack:
@geealbers I think you are right, it looks like `urlFilter` was doubling the subpaths. I'm assuming we were using that for validation/sanitation purposes?

Let me know if this works


geealbers (approved):
Yep, that did it. I deployed the fix to French Silver and everything's working now, and I also tested it on a separate demo project. Thanks!

PR #706, opened 2023-04-07
Refactor search component and fix index path

hbalenda:
Changes:
- Replace `search.liquid` with `search.js` component, this was the last remaining liquid template in the base layout, seems easier to maintain all as js
- Update search index path to handle subpath


packages/11ty/_includes/components/search.js
```diff
@@ -0,0 +1,61 @@
...
+                          <!-- js-search-results-template -->
+                      </ul>
+                      </div>
+                  </div>
+              </section>
+          </div>
+      </div>
+    `
+  }
+}
```
hbalenda, line 61:
```suggestion
}

```

mphstudios (approved):

PR #713, opened 2023-04-19
Update `quire create --quire` option to accept local package

hbalenda:
Changes:
- Refactors `--quire` version option to accept a local path to a quire 11ty package
- Updates the 11ty `package-lock.json` which was out of date

### DIFF
--- a/packages/cli/src/lib/quire/index.js
+++ b/packages/cli/src/lib/quire/index.js
@@ -200,7 +200,8 @@ async function install(options = {}) {
  */
 async function installInProject(projectPath, options = {}) {
   const version = options.quire || 'latest'
-  console.debug(`[CLI:quire] installing quire-11ty@${version} into ${projectPath}`)
+  const quire11tyPackage = fs.existsSync(options.quire) ? options.quire : `${PACKAGE_NAME}@${version}`
+  console.debug(`[CLI:quire] installing ${quire11tyPackage} into ${projectPath}`)
 
   /**
    * Delete the starter project package configuration so that it can be replaced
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/quire/index.js
```diff
@@ -200,7 +200,8 @@ async function install(options = {}) {
  */
 async function installInProject(projectPath, options = {}) {
   const version = options.quire || 'latest'
-  console.debug(`[CLI:quire] installing quire-11ty@${version} into ${projectPath}`)
+  const quire11tyPackage = fs.existsSync(options.quire) ? options.quire : `${PACKAGE_NAME}@${version}`
+  console.debug(`[CLI:quire] installing ${quire11tyPackage} into ${projectPath}`)
```
mphstudios, line 204: @hbalenda being able to easily install a local version is going to be very helpful. 

We may want to make the logic for determining wether to install `quire-11ty` from a local package or published version more explicit by using separate variables for a local package or version option, something like below
```
const localQuire11ty = fs.existsSync(options.quire) || null
const quireVersion =  localQuire11ty ? getQuire11tyVersion(localQuire11ty) : options.quire
const quire11tyPackage = localQuire11ty || `${PACKAGE_NAME}@${version}`
```

or by defining these as two explicitly distinct command options (see [Commander `options` configuration](https://github.com/tj/commander.js#more-configuration) for `conflicts` and `implies`,

```
    options: [
       [ '--quire-path <path>', 'local path to quire-11ty version to install' ],
       [ '--quire-version <version>', 'quire-11ty version to install', 'latest' ],
    ],
```

We could handle the conflicting option here in the code though Command provides an `addOption` that could be used with some refactoring of how we read the command configuration to programmatically build options.

mphstudios (commented):
@hbalenda see comments in the code thinking about how we can implement this option be most explicit and perhaps easier to maintain if we need to change the logic; this feature is going to be very useful for development.

packages/cli/src/lib/quire/index.js
mphstudios: If we choose to keep a single `--quire` options we should see if using a [custom option processing function](https://github.com/tj/commander.js#custom-option-processing) has any benefit. Using a custom option processor would encapsulate the logic to determine if the user passed a path or version string and display display specific error messaging if the path or version is not valid.

hbalenda:
@mphstudios I think having two options `--quire-version` and `--quire-path` makes sense, I've made that change.

mphstudios (approved):

hbalenda:
PARTY!

PR #714, opened 2023-04-19
Include eleventyComputed and application javascript from starter

hbalenda:
Changes:
- Moves application javascript and eleventyComputed from `quire-starter-default` into `quire/packages/11ty/`. These are tightly coupled with the components and should live here for a simpler development process.
- Removes `content` directory from `.gitignore`


mphstudios (commented):
@hbalenda what are your thoughts about moving these files into `quire-11ty/assets/` and `quire-11ty/computed/` directories (instead of excluding `content/`) then opening a corresponding pull-request for `quire-11ty` that copies them into the project content directory.

hbalenda:
@mphstudios I think leaving them in `content` makes it clear what those files are, and that the directory contents will be merged with `starter/content`

mphstudios (approved):

PR #721, opened 2023-05-02
Update the URLs used for videos in PDF and EPUB output

geealbers:

### Checklist 

- [X] I have read the [CONTRIBUTING.md](https://github.com/thegetty/quire/blob/main/CONTRIBUTING.md) file

n/a

To update the Vimeo and YouTube URLs that are added above the figure captions in the PDF and EPUB output, so that they work consistently and are in as short a form as possible for readers.

Previously, the emitted Vimeo URL was the same as the embed URL and not consistently working, and the YouTube URL worked but there's a shorter version available. So I updated the `sourceUrl` for each.

One question I had was around the `replace()` function being used for the Vimeo URLs. It's necessary and working, but it assumes the `media_id` will be supplied as a string. In a case like `media_id: 672853278` where the id is not quoted, it is read as an integer and breaks the build/preview process when it runs into `replace()`. Could/should we also be converting the `media_id` to a string to eliminate this error? Or is this just something we need to document for users? To tell them they always need to quote `media_id`s?

n/a

n/a

### DIFF
--- a/packages/11ty/_includes/components/figure/media-embed-url.js
+++ b/packages/11ty/_includes/components/figure/media-embed-url.js
@@ -33,16 +33,18 @@ module.exports = function (eleventyConfig) {
     },
     vimeo(mediaId) {
       const baseUrl = 'https://player.vimeo.com/video/'
+      const sourceBaseUrl = 'https://vimeo.com/'
       // Sample Vimeo id: 672853278/b3f8d29d53
+      // or id: 672853278
       const embedId = mediaId.replace('/', '?h=')
       return {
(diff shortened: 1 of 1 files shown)


geealbers:
@thegetty/int-dev I made this small fix in our Getty books, so I thought I'd submit it as a PR here. (Though granted, I maybe should have created an issue for it first?) 

I posed a question above about converting the `media_id` to a string. Can you take a look whenever you have a chance and let me know what you think? No rush.

hbalenda (approved):

PR #727, opened 2023-05-15
DEV-14988: Add extra height constraints to prevent image-sequences from rendering outside of catalog lightbox boundaries

anderspollack:
See these companion PRs with lightbox style updates (all three are identical, we should really move these styles into quire core soon):
- https://github.com/thegetty/quire-starter-default/pull/30
- https://github.com/thegetty/quire-starter-iiif/pull/3
- https://github.com/thegetty/quire-starter-demo/pull/1

### DIFF
--- a/packages/11ty/_includes/components/figure/image/image-sequence.webc
+++ b/packages/11ty/_includes/components/figure/image/image-sequence.webc
@@ -415,7 +415,13 @@ img.${placeholder} {
 }
 </script>
 <style>
+image-sequence {
+  display: block;
+  height: 100%;
+}
+
 .image-sequence.interactive {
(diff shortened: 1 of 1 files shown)


geealbers (approved):
That did it! I made your changes in a sample project and tested on Chrome, Safari, Firefox, and Edge. All looked good. Will approve here and in the other content repos. Thanks!

geealbers:
As a side note on those three content repos, as far as I know both @hbalenda and I have failed to update the `iiif` and `demo` ones as we've made changes to `quire-starter-default` for accordions and PDF output. Another thing to consider as we're also talking about versioning and where the content stuff lives, etc.

PR #738, opened 2023-05-18
DEV-14841: Ensure annoref links don't link to an `undefined` image sequence index

anderspollack:
If the `index` was not specified in an annoref shortcode, it was passing `undefined` as an index to the `<image-sequence>`. Now, if an `index` is omitted, the annoref shortcode links to a specifed figure `startCanvasIndex`, or `0` if the figure does not have a `startCanvasIndex`


geealbers:
This works except seems to have an adverse on the error messaging. Prior to the change, if you used the annoref shortcode with a figure id that didn't exist, you'd get the following error. Which is not structured in the way we're doing currently with the `logger` but is otherwise correct:

```
[annoref shortcode] "fig" parameter doesn't correspond to a valid figure id in "figures.yaml". Fig: rotating-test
```

After the change, the error messaging is:

```
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble rendering liquid template ./content/all.md (via TemplateContentRenderError)
[11ty] 2. Cannot destructure property 'startCanvasIndex' of 'figure' as it is undefined., file:./content/all.md, line:7, col:1 (via RenderError)
[11ty] 3. Cannot destructure property 'startCanvasIndex' of 'figure' as it is undefined. (via TypeError)
[11ty] 
[11ty] Original error stack trace: TypeError: Cannot destructure property 'startCanvasIndex' of 'figure' as it is undefined.
[11ty]     at /Users/galbers/GitHub/in-sequence/_plugins/shortcodes/annoref.js:21:13
[11ty]     at renderComponent (/Users/galbers/GitHub/in-sequence/_plugins/components/addComponentTag.js:34:37)
[11ty]     at Tag.render (/Users/galbers/GitHub/in-sequence/_plugins/components/addComponentTag.js:49:22)
```

anderspollack:
@geealbers I fixed the bad error message, now it should be back to the helpful one when passing an invalid figure id. This also updates the image sequence `index` to be stored as a query parameter. Ready for review!

geealbers (approved):
Tested and approved!

PR #760, opened 2023-06-21
DEV-15269: Render static image tag for sequence figures inline

anderspollack:
This update renders image sequence figures as a static `` tag inline, retaining interactive click + drag functionality when opened in a modal.

@geealbers I hope this solves a good chunk of french-silver performance problems, let me know if this change is working

### DIFF
--- a/packages/11ty/_includes/components/figure/image/image-tag.js
+++ b/packages/11ty/_includes/components/figure/image/image-tag.js
@@ -13,8 +13,8 @@ const path = require('path')
 module.exports = function(eleventyConfig) {
   const { imageDir } = eleventyConfig.globalData.config.figures
 
-  return function ({ alt='', src='' }) {
-    const imageSrc = src.startsWith('http') ? src : path.join(imageDir, src)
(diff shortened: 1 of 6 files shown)


packages/11ty/_includes/components/figure/image/image-tag.js
```diff
@@ -13,8 +13,8 @@ const path = require('path')
 module.exports = function(eleventyConfig) {
   const { imageDir } = eleventyConfig.globalData.config.figures
 
-  return function ({ alt='', src='' }) {
-    const imageSrc = src.startsWith('http') ? src : path.join(imageDir, src)
+  return function ({ alt='', src='', isStatic=false }) {
+    const imageSrc = src.startsWith('http') || isStatic ? src : path.join(imageDir, src)
```
anderspollack, line 17: I don't love calling this option `isStatic` but I need some kind of flag to omit the `imageDir`, since static outputs (`print-image`, `thumbnail`, and `static-figure-image` all live in `_site/iiif`, rather than `_assets/images`)

PR #762, opened 2023-06-22
Feature: starter-11ty compatibility

hbalenda:
Changes:
- Refactor `latest` to accept a semantic version range, such as `^1.0.0-rc`
- Log error message and clean up project if init starter fails
- Remove `latest` as default quire version option value. quire-11ty version will be determined by the option if provided or default to the version required by the starter.

### DIFF
--- a/packages/cli/src/commands/create.js
+++ b/packages/cli/src/commands/create.js
@@ -23,7 +23,7 @@ export default class CreateCommand extends Command {
     ],
     options: [
       [ '--quire-path <path>', 'local path to quire-11ty package' ],
-      [ '--quire-version <version>', 'quire-11ty version to install', 'latest' ],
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/create.js
```diff
@@ -23,7 +24,7 @@ export default class CreateCommand extends Command {
     ],
     options: [
       [ '--quire-path <path>', 'local path to quire-11ty package' ],
-      [ '--quire-version <version>', 'quire-11ty version to install', 'latest' ],
+      [ '--quire-version <version>', 'quire-11ty version to install' ],
       // [ '--eject', 'install quire-11ty into the project directory', true ],
```
mphstudios, line 28:
```suggestion
```

mphstudios (approved):
Very nice.

PR #775, opened 2023-07-06
DEV-15497: Fix failing figure unit tests; add sequence `files` fixtures

anderspollack:
While publishing new `rc` versions of `quire-11ty`, the publish process was not completing due to failing tests. This fixes failing figure plugin unit tests. These tests were originally depending on a specifically-named test publication to exist on a user's machine.

This refactor sequence factory `create()` method and `getSequenceFiles()` helper function to support passing an array of `files` directly to circumvent the directory lookup functionality in `getSequenceFiles()`.

```sh
cd packages/11ty
npm run test
```


packages/11ty/_plugins/figures/test/__fixtures__/iiif-config.json
```diff
@@ -2,9 +2,9 @@
   "baseURI": "http://localhost:8080",
   "dirs": {
     "imagesDir": "_assets/images",
-    "inputRoot": "/Users/apollack/Code/getty/sequence-blarg/content",
+    "inputRoot": "/test-project/_site",
     "outputPath": "iiif",
-    "outputRoot": "/Users/apollack/Code/getty/sequence-blarg/_site"
+    "outputRoot": "/test-project/_site"
```
anderspollack, line 7: the `inputRoot` and `outputRoot` are not actually being used as these generated figures are coming from fixture data, but it seems silly to have a fake path on my personal machine in this JSON...

packages/11ty/_plugins/figures/sequence/index.js
```diff
@@ -12,12 +12,12 @@ const Annotation = require('../annotation')
...
  */
 module.exports = class Sequence {
-  constructor(figure, sequence) {
+  constructor(figure, sequence, files) {
     const { iiifConfig } = figure
     this.behavior = sequence.behavior
     this.dir = sequence.id
     this.figure = figure
-    this.files = getSequenceFiles(sequence, iiifConfig)
+    this.files = getSequenceFiles(sequence, iiifConfig, files)
```
mphstudios, line 20: @anderspollack this makes sense and if we make gathering the `files` a concern of the factory then it can pass them to the `Sequence` constructor, which may make both easier to test with a fixture for the files.

mphstudios:
@anderspollack does `getSequenceFiles` need to be a separate module or should it be a method of the factory?

packages/11ty/_plugins/figures/sequence/factory.js
```diff
@@ -13,11 +15,40 @@ module.exports = class SequenceFactory {
...
+    const sequenceDir = path.join(inputRoot, imagesDir, sequence.id)
+    const defaultSequenceRegex = /^\d{3}\.(jpg|png)$/
+    const sequenceRegex = sequence.regex
+      ? new RegExp(sequence.regex.slice(1, -1))
+      : defaultSequenceRegex
+    if (!fs.existsSync(sequenceDir)) return
+    return fs
+      .readdirSync(sequenceDir)
+      .filter((sequenceItemFilename) => sequenceItemFilename.match(sequenceRegex))
   }
```
mphstudios, line 53: @anderspollack since the `iiifConfig` is passed to the `sequenceFactory` as an argument it probably not needed as an argument to `getSequenceFiles`, a few additional comments to this method inline below—

```suggestion
  getSequenceFiles(sequence) {
    if (!sequence) return
    
    const sequenceDir = path.join(inputRoot, imagesDir, sequence.id)
    
    /**
     * Default file name pattern for 360° images,
     * zero padded three digit degrees
     * @todo refactor to use default pattern in publication configuration
     */
    const defaultPattern = /^\d{3}\.(jpg|png)$/
    
    const sequenceRegex = sequence.regex
      ? new RegExp(sequence.regex.slice(1, -1))
      : defaultPattern
    
    // @todo log a warning or perhaps throw an error?
    if (!fs.existsSync(sequenceDir)) return
    
    return fs
      .readdirSync(sequenceDir)
      .filter((filename) => filename.match(sequenceRegex))
  }
```

mphstudios (commented):
@anderspollack see comments in line with the `getSequenceFiles` method.

mphstudios (approved):

PR #776, opened 2023-07-10
Render static figures inline; hide annotations inline

anderspollack:
**Work-in-progress**

- Render `staticInlineFigure` images inline, only show tiled IIIF images in lightbox / modal. Because the inline figures are just `` tags, the annotation UI should also be shown only in lightbox / modal

This branching logic in `figure/image/element.js` could probably be made a lot more concise, but this works at the moment and yields some pretty immediate performance gains.

### DIFF
--- a/packages/11ty/_includes/components/figure/image/element.js
+++ b/packages/11ty/_includes/components/figure/image/element.js
@@ -28,9 +28,17 @@ module.exports = function (eleventyConfig) {
           return await imageSequence(figure, options)
         }
       case isCanvas:
-        return canvasPanel(figure)
+        if (!interactive && staticInlineFigureImage) {
+          return imageTag({ alt, src: staticInlineFigureImage, isStatic: !interactive })
+        } else {
+          return canvasPanel(figure)
+        }
       case isImageService:
-        return imageService(figure)
+        if (!interactive && staticInlineFigureImage) {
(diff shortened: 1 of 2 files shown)


mphstudios (approved):

geealbers:
@anderspollack From my initial tests with _French Silver_, this really makes a significant difference especially with all those pages that are heavy in zooming images. I don't think it works from a user/reader standpoint to also make the checkbox and radio images static inline though, at least not by default. Too much is lost in those situations (detailed below). Could we try not rendering them static inline at all and just leaving them as they were? 

The other option would be adding a config value that allows users to choose whether or not to render static versions inline. Though in that case I think we need to also then add an icon to them like with the rotating images, and we need to ensure that whatever images are marked as `selected: true` are rendered and stacked inline.

These are the issues I'm seeing when annotation images are static inline:

- there is no indication that these images have annotations/layers at all and so might be worth opening in the modal
- the checkbox and radio labels often act as a kind of key or legend for the image giving readers more information about what the image is, so when they are missing the image makes a little less sense
- Quire users can specify what layers are selected when the page first loads, but this becomes less valuable if it's just visible in the modal
- For checkbox images, only the base image is rendering, even if other layers are `selected: true` in the yaml and this is confusing.

And fwiw, radio button images are still using canvas panel from what I can tell, rather that coming out as simple `` tags like checkbox images are.

PR #786, opened 2023-07-19
Sort epub reading order by url

hbalenda:
Changes:
- Adds `sortByKeys` filter
- Sorts epub reading order by `url` key

### DIFF
--- a/packages/11ty/_plugins/filters/sortByKeys.js
+++ b/packages/11ty/_plugins/filters/sortByKeys.js
@@ -26,7 +26,6 @@ module.exports = (sortKeys = [], sortOrder = 'ascending') => {
 
   if (!['ascending', 'descending'].includes(sortOrder)) {
     throw new Error(`
-
       Invalid sort order "${sortOrder}" passed to sortByKeys,\n
       value of sortOrder argument must be either "ascending" or "descending".
     `)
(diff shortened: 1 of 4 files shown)


packages/11ty/_plugins/filters/sortByKeys.js
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @return  {Number}  Comparison 1, 0, -1
+ */
+module.exports = (sortKeys = [], sortOrder = 'ascending') => {
+  if (!sortKeys.length) return
+
+  if (!['ascending', 'descending'].includes(sortOrder)) {
+    throw new Error(`
+
+      Invalid sort order "${sortOrder}" passed to sortByKeys,\n
```
mphstudios, line 30: If the additional empty line is intentional we might instead use a new line character, `\n`, like so
```suggestion
    throw new Error(`\n
      Invalid sort order "${sortOrder}" passed to sortByKeys,\n
```

mphstudios (dismissed):

packages/11ty/_plugins/filters/sortByKeys.js
```diff
@@ -0,0 +1,119 @@
...
+ * @param  {String}  sortOrder - Default sort order, "ascending" or "descending"
+ *
+ * @return  {Number}  Comparison 1, 0, -1
+ */
+module.exports = (sortKeys = [], sortOrder = 'ascending') => {
+  if (!sortKeys.length) return
+
+  if (!['ascending', 'descending'].includes(sortOrder)) {
+    throw new Error(`
+
```
hbalenda, line 29:
```suggestion
```

PR #788, opened 2023-07-20
DEV-15455: Replace all `&nbsp;` and `&#32;` entities with unicode ASCII spaces `\u0020`

anderspollack:
This update removes all HTML entities in layouts and shortcodes. All entities were originally meant to ensure intentional whitespace was preserved inside tags for screen readers, but since they are not allowed in EPUB, they have been replaced with unicode ASCII space characters (`\u0020`)

### DIFF
--- a/packages/11ty/_layouts/cover.liquid
+++ b/packages/11ty/_layouts/cover.liquid
@@ -11,7 +11,8 @@ layout: base.11ty.js
   <div
     class="quire-cover__overlay"
     style="background-image: url('{{ imagePath }}');"
-  />
+  >
+  </div>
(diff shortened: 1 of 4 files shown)


packages/11ty/_layouts/cover.liquid
```diff
@@ -11,7 +11,8 @@ layout: base.11ty.js
   <div
     class="quire-cover__overlay"
     style="background-image: url('{{ imagePath }}');"
-  />
+  >
+  </div>
```
anderspollack, line 15: this self-closing tag was also inserting a bunch of unnecessary HTML space entities in the EPUB output

mphstudios (approved):

PR #800, opened 2023-07-26
DEV-15479: Fix `dateModified` not formatting/validating correctly in EPUB manifest

anderspollack:
- Replace misspelled `dateModifed` key with `dateModifed` in epub manifest
- Strip milliseconds (`.sss`) from ISO date string in EPUB manifest `dateModified` field

### DIFF
--- a/packages/11ty/_plugins/transforms/outputs/epub/manifest.js
+++ b/packages/11ty/_plugins/transforms/outputs/epub/manifest.js
@@ -118,6 +118,13 @@ module.exports = (eleventyConfig) => {
     ? removeHTML(full).replace(/\r?\n|\r/g, ' ')
     : oneLine
 
+  /**
+   * Strip milliseconds from ISO date string (`.sss`)
+   */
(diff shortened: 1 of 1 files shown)


packages/11ty/_plugins/transforms/outputs/epub/manifest.js
```diff
@@ -118,6 +118,13 @@ module.exports = (eleventyConfig) => {
     ? removeHTML(full).replace(/\r?\n|\r/g, ' ')
     : oneLine
 
+  /**
+   * Strip milliseconds from ISO date string (`.sss`)
+   */
+  const pubDateWithoutMs = pubDate
+    .toISOString()
+    .replace(/\.\d{3}/, '')
```
anderspollack, line 126: I know this could be done more expressively with something like `moment.js` but I wasn't sure if we wanted to add a dependency for just one thing

mphstudios (approved):

Issue #752, opened 2023-06-09
Pages or directories at the root level that start with a number break the pdf generation


geealbers:
### Before proceeding, check to make sure there isn’t an existing issue for this bug.

- [X] I have searched the existing issues and determined this is a new bug.

### Steps to reproduce

1. Start a new project
2. Rename a directory and/or a markdown file in the `content` directory so that it starts with a number (0–9).
3. run `quire build` and then `quire pdf`

### Actual behavior

An error is thrown and the pdf is not generated

```
Error: Evaluation failed: DOMException: Failed to execute 'querySelector' on 'Element': '#10-intro' is not a valid selector.
```

### Expected behavior

The PDF would generate and no error would be thrown.

Sub-directories and markdown pages within directories can have file names that start with numbers and not break the PDF. So the actual behavior is inconsistent. It's also problematic that the error doesn't happen until the user tries to generate the PDF as this is often late into the development process. Also that the error messaging isn't helpful unless one happens to know `id` selectors can't start with numbers.

### Version numbers

macOS Ventura 13.2.1 (22D68)
Node 18.16.0
Quire CLI 1.0.0-rc.7
Quire 11ty 1.0.0-rc.10

Erin-Cecele:
We have confirmed that the reported issue is a bug. It has been added to our development queue. Stay tuned for updates.

Erin-Cecele:
Our developers are currently working on this issue. Please check back for updates.

Erin-Cecele:
This was fixed in PR #808. Issue is now closed.

PR #808, opened 2023-08-03
Fix/GH-752 content pages starting with numbers break pdf build

aprigge:

### Checklist 

- [x] I have read the [CONTRIBUTING.md](https://github.com/thegetty/quire/blob/main/CONTRIBUTING.md) file

- [x] I have made my changes in a new branch and not directly in the main branch

 #752 

The PDF would generate and no error would be thrown when a page or directory within `content` begins with a number.

Prepend `page` to selectors derived from files within the `content` directory.

n/a

n/a

### Additional Comments
n/a


packages/11ty/_layouts/base.11ty.js
```diff
@@ -10,7 +10,7 @@ const { html } = require('~lib/common-tags')
 module.exports = async function(data) {
   const { pageClasses, collections, content, pageData, publication } = data
   const { inputPath, outputPath, url } = pageData || {}
-  const pageId = this.slugify(url) || path.parse(inputPath).name
+  const pageId = `page_${this.slugify(url)}` || `page_${path.parse(inputPath).name}`
```
hbalenda, line 13: This statement will always return the first option, if `this.slugify(url)` is undefined the `pageId` will be `page_` rather than `page_${path.parse(inputPath).name}`. Instead you can do something like,
```suggestion
  const pageId = `page_${this.slugify(url) || path.parse(inputPath).name}`
```
or 

```
const id = this.slugify(url) || path.parse(inputPath).name
const pageId = `page_${id}`
```

hbalenda: Also, we tend to use slugify with ids, so kebab-case (`page-*`) is preferable here and in the transform to snake case.

packages/11ty/_plugins/transforms/outputs/pdf/transform.js
```diff
@@ -82,7 +82,8 @@ module.exports = function(eleventyConfig, collections, content) {
         setDataAttributes(currentPage, sectionElement)
 
         // set an id for anchor links to each section
-        sectionElement.setAttribute('id', mainElement.dataset.pageId)
+        const computedPageID = `page_${mainElement.dataset.pageId}`
+        sectionElement.setAttribute('id', computedPageID)
```
hbalenda, line 86: The value of `mainElement.dataset.pageId` already has the page-prefixed since it is pulled from the template generated by `base.11ty.js`, so you don't need to re-compute it here
```suggestion
        sectionElement.setAttribute('id', mainElement.dataset.pageId)
```

hbalenda (changes requested):

packages/11ty/_plugins/transforms/outputs/pdf/transform.js
aprigge: Thanks @hbalenda , I've made these changes.

hbalenda (approved):
🎉 🎉 🎉

PR #813, opened 2023-08-09
DEV-15458: Ensure background image assets are copied into EPUB package

anderspollack:
This seemed like the most straightforward way to ensure that assets *only* referenced in inline style `background-image` declarations get copied into EPUBs. I can see some potential value in adding more programmatic logic to the EPUB transform to query select all things with `style*=background-image` and writing a hidden image tag inside that element, but that approach sounds pretty brittle and makes assumptions about how users write layout markup/styles. 

Not sure if this kind of thing also warrants extra documentation, and if so, where that documentation might go.


hbalenda:
This fixes the errors, but I noticed with the cover image, the static cover image that's used as the cover for the epub is already being copied in, but the error comes from the _other_ cover image, which is still included in the markup as the `background-image` but not needed. Also I don't think the splash image is being rendered in the epub even now that it's included in the assets, unless I'm mistaken, it might need to be included as you have it and not `visually-hidden`, and `data-outputs-include="epub"`  only?

hbalenda (approved):

PR #814, opened 2023-08-09
Add screen-only stylesheet

hbalenda:
https://github.com/thegetty/quire-starter-default/pull/36 and https://github.com/thegetty/quire-starter-iiif/pull/9 should be merged first. 

Since pagedjs reads styles with the `@media screen` media query and does not allow overriding `position: fixed`, we need to include a separate stylesheet that is only included in the site output. 

See pagedjs issue https://gitlab.coko.foundation/pagedjs/pagedjs/-/issues/433

### DIFF
--- a/packages/11ty/CHANGELOG.md
+++ b/packages/11ty/CHANGELOG.md
@@ -23,6 +23,7 @@ Changelog entries are classified using the following labels:
 ### Added
 
 - Adds print `table` component without a modal link
+- Import `screen.scss` into `javascript/application/index.js`
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


anderspollack:
I think this totally works for now! I was trying to adjust the `.quire-primary` flex container and make the `.quire-navbar` fixed instead of `position: sticky` to remove the need for `.quire-entry__content` to be `position: fixed`, but that was affecting other layout styles in various small ways (like requiring some extra calculated margins) and it seems much better to not accidentally introduce regressions... at least until we can do the big style refactor.

anderspollack (approved):

PR #816, opened 2023-08-09
Page class inheritance

hbalenda:
Changes:
- Renames `eleventyComputed` property `pageClasses` and frontmatter property `class` to `classes` for consistency, avoiding using the reserved word `class`, and to handle merging template, layout, and computed classes.

Related to https://github.com/thegetty/quire-starter-default/pull/37


geealbers (dismissed):
Looks like a good solution. Thanks @hbalenda.

@Erin-Cecele, if you haven't already, let's be sure to add an issue to the `quire-docs` repo for us update the documentation for this. Rather than using this syntax to add custom page classes:

```yaml
class: page-one fancy-custom-class
```
Users will use:

```yaml
classes:
  - page-one
  - fancy-custom-class
```

PR #832, opened 2023-08-21
Update print image transformation width

hbalenda:
Updates the `print-image` transformation width to 2025px and sets option to not enlarge images < 2025px.


anderspollack (approved):

geealbers:
@hbalenda doesn't this also need to be changed in `_plugins/figures/iiif/config.js`?

PR #834, opened 2023-08-22
Remove numbers from epub filenames

hbalenda:
Changes:
- Refactors epub reading order to use a sort key rather than prepending the index and sorting by filename. Fixes validation errors for ids that begin with numbers.

### DIFF
--- a/packages/11ty/CHANGELOG.md
+++ b/packages/11ty/CHANGELOG.md
@@ -24,13 +24,17 @@ Changelog entries are classified using the following labels:
 
 - Validation method to check if an image can be tiled, and log error if invalid.
 
-### Removed
-
-- Remove `.jp2` from supported image extensions
-
 ### Changed
 
 - Increase `print-image` transformation width to `2025px`
+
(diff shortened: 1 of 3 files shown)


hbalenda:
@geealbers good catch, I added the `page-` prefix here as well

packages/11ty/CHANGELOG.md
```diff
@@ -24,14 +24,18 @@ Changelog entries are classified using the following labels:
...
-
-- Remove `.jp2` from supported image extensions
-
 ### Changed
 
 - Increase `print-image` transformation width to `2025px`
 
+### Fixed
+
+- Sort epub pages for reading order by sort key rather than filename. Fixes epub validation error for filenames that begin with numbers.
```
hbalenda, line 33:
```suggestion
- Sort epub pages for reading order by sort key rather than filename. Fixes epub validation error for filenames that begin with numbers.
- Prefix epub filename with `page-` to ensure validity if filename begins with a number
```

PR #837, opened 2023-08-22
Relative paths in epub output

hbalenda:
Changes:
- Remove preceding slash from relative paths in epub output, which cause validation errors

### DIFF
--- a/packages/11ty/_plugins/transforms/outputs/epub/transform.js
+++ b/packages/11ty/_plugins/transforms/outputs/epub/transform.js
@@ -42,6 +42,29 @@ module.exports = function(eleventyConfig, collections, content) {
   }
 
   /**
+   * Removes preceding slashes from asset paths
+   * @param {HTMLElement} element 
+   */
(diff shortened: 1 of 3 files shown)


packages/11ty/_plugins/transforms/outputs/epub/transform.js
```diff
@@ -41,6 +41,29 @@ module.exports = function(eleventyConfig, collections, content) {
...
+    images.forEach((item) => {
+      const src = item.getAttribute('src')
+      if (!src) return
+      item.setAttribute('src', src.replace(/^\//, ''))
+    })
+    links.forEach((item) => {
+      const href = item.getAttribute('href')
+      if (!href) return
+      item.setAttribute('href', href.replace(/^\//, ''))
+    })
```
anderspollack, line 64: wondering if we should also handle `<video>` tags here, as they are supported in EPUB3?

anderspollack (approved):

anderspollack (approved):

PR #850, opened 2023-10-03
DEV-16146: Render all attributes in object cards; specify order of attributes in `objects.yaml`

anderspollack:
This update allows an `objects-page` layout to render object cards with _all_ object attributes. By default, an object card displays every attribute. Field order can be specified in `objects.yaml`: `object_card` property:
```yaml
object_card:
  - thumbnail
  - name
``` 
This also refactors our components to use `<script webc:setup>` block for more consistency


packages/11ty/_includes/components/object-card.webc
```diff
@@ -56,10 +65,10 @@ const renderFields = () => {
...
 }
 
-.object-card__image {
-  padding: 1.5em 1.5em 0em 1.5em;
-}
-
+/* .object-card__image {
+   wid
+   }
+ */
```
anderspollack, line 71:
```suggestion
```

geealbers:
@anderspollack I'm seeing three things, listed below. And I just submitted a PR to `quire-starter-objects-test` with the additional data I created in order to test these.

**1.**

If one of the objects is missing any of the keys listed under `object_card`, this error is thrown and preview stops:

```
[11ty] 1. Having trouble writing to "_site/things/index.html" from "./content/things/index.md" (via EleventyTemplateError)
[11ty] 2. Check the webc:type="js" element in ./_layouts/objects-page.webc.
[11ty] Original error message: Cannot read properties of undefined (reading 'split') (via Error)
[11ty] 
[11ty] Original error stack trace: Error: Check the webc:type="js" element in ./_layouts/objects-page.webc.
[11ty] Original error message: Cannot read properties of undefined (reading 'split')
[11ty]     at ModuleScript.evaluateScript (/Users/galbers/test-object-filters/node_modules/@11ty/webc/src/moduleScript.cjs:71:10)
[11ty]     at Object.<anonymous> (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:81:41)
[11ty]     at AstSerializer.transformContent (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:515:42)
[11ty]     at AstSerializer.compileNode (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:1044:20)
[11ty]     at AstSerializer.getChildContent (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:349:67)
[11ty]     at AstSerializer.compileNode (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:1221:46)
[11ty]     at async AstSerializer.getChildContent (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:349:56)
[11ty]     at async AstSerializer.compileNode (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:1226:36)
[11ty]     at async AstSerializer.getChildContent (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:349:56)
[11ty]     at async AstSerializer.compileNode (file:///Users/galbers/test-object-filters/node_modules/@11ty/webc/src/ast.js:1226:36)
[11ty] Benchmark    544ms  32%   124× (Configuration) "citation" JavaScript Function
[11ty] Benchmark    589ms  34%    62× (Configuration) "menu" JavaScript Function
[11ty] Benchmark    542ms  31%   124× (Configuration) "formatCitation" JavaScript Function
```

**2.**

It looks like the key that highest up in the source object data is the one that's ending up with the `class="object-card__title">` and so is bold and red. Instead, it should be the key that is highest up (first) in the `object_card` list that should get this class.

**3.**

What do you think about including the names of the keys from the `object_card` list on the html elements in someway, in order to facilitate style customization down the road? My first thought was to add them as class names, but I wondered if someone might inadvertently use a key that happens to already have styles associated with it. So, then thought maybe a `data` attribute? Not sure what the best way to do this would be.

And no worries if this is difficult somehow, or if you just don't think it's a good idea. Folks could also just use an nth-child selector to style these individually.

```html
<div class="object-card__content">
  <span class="object-card__field" data-object-attr="letter"></span>
  <figure class="object-card__image" data-object-attr="thumbnail"></figure>
  <span class="object-card__field" data-object-attr="name"></span>
  <span class="object-card__field" data-object-attr="artist"></span>
  <span class="object-card__field" data-object-attr="time_span"></span>
</div>
```

anderspollack:
@geealbers all of your issues / requests should be addressed
1. fields in `objects.yaml` > `object_card` that are not defined on an object will be ignored
2. order of `objects.yaml` > `object_card` should be respected at the moment, the first _text_(non-thumbnail) element should be receiving the bold/arrow icon treatment
3. I added the attribute field names to a `data-field-name` attribute on each field element in the object card

geealbers:
@anderspollack In my test, it's showing the first item defined under `object_card` and the `thumbnail` if it's defined there as well, but nothing else.

And just to confirm, I am seeing the `data-field-name` so I have the latest commit.

anderspollack:
@geealbers ah this should be fixed now, I hadn't tried it with the updated starter yet but it should be rendering all fields specified after the thumbnail

packages/11ty/_includes/components/object-card.webc
```diff
@@ -1,48 +1,73 @@
...
+      ? thumbnail.src
+      : thumbnail
+
+    /**
+     * If no IIIF images, point to the raw assets directory
+     */
+    return src && !src.startsWith('/iiif')
+      ? `${this.config.figures.imageDir}/${src}`
+      : src;
+  }
```
anderspollack, line 33: I moved all of this thumbnail `src`-mangling logic from `objects-catalog` to here to clean that component up

packages/11ty/_includes/components/object-card.webc
```diff
@@ -1,48 +1,73 @@
...
+
+<script webc:setup>
+const renderFields = (objectCardAttributes) => {
+  const { $data, figures } = objectCardAttributes;
+  const ignoredAttributes = ['uid', 'url'];
+  const isIgnoredAttribute = (key) => !ignoredAttributes.includes(key);
+  const attributeOrder = $data.object_card || $data.objects.object_card || Object.keys(objectCardAttributes).filter(isIgnoredAttribute);
+  const attributes = attributeOrder.map((key) => ({ name: key, value: objectCardAttributes[key]}));
+  const firstTextAttribute = attributes.find(({ name }) => name !== 'thumbnail');
+  const alt = () => typeof thumbnail === 'object' ? thumbnail.alt : '';
```
anderspollack, line 13: this adds support for `thumbnail` as an object to allow users to define an `alt` for the object image

geealbers:
@anderspollack, the fields are rendering but the `object-card` links are all `href="this.webc.attributes.url"`.

Also, the thumbnails are getting `alt="alt(this.webc.attributes)"` added to them. And since we can't add alt text to the thumbnails through yaml, we should actually leave these explicitly blank (`alt=""`) so that screen readers ignore them altogether. And we can document that for users.

Finally, wanted to point out that the HTML elements (`<figure class="object-card__image">` and `<span class="object-card__field">`) are being created on the card even they are empty of content. Was that the intention? Fine if so, just wanted to make sure.

packages/11ty/_includes/components/object-filters.webc
```diff
@@ -1,37 +1,16 @@
...
+    <div class="object-filters__select">
+      <select :id="filter.id" data-control-filter :data-control-name="filter.fieldName">
+        <option value="">All</option>
+        <option webc:for="option of filter.options" :value="option.value" @text="option.label"></option>;
+      </select>
+    </div>
+  </div>
+</div>
+
+<script webc:setup>
```
mphstudios, line 13:
```
/**
 * Capitalize the first letter of a string
 */
const capitalizeFirstLetter = (string) => string[0].toUpperCase() + string.slice(1);
```

Using `slugify` is uncessary as `String.prototype.split` accepts a Regex
```
/**
 * Convert snake_case or kebab-case strings to title case
 */
const humanize = (string) => {
  return string
    .split(/[_-]/)
    .map(capitalizeFirstLetter)
    .join(' ');
}
```

packages/11ty/_includes/components/object-filters.webc
```diff
@@ -53,48 +32,69 @@ const humanize = (string) => {
...
+
+        values.forEach((value) => {
+          if (value && !existingValues.includes(value)) {
+            uniqueOptions.push({ label: value, value });
+          }
+        })
+
+        return uniqueOptions;
+      }, [])
+      .sort((a, b) => a.value > b.value ? 1 : -1)
```
mphstudios, line 85: Alternate approach destructing `value` as `a` and `b`
```suggestion
      .sort(({ value: a}, { value: b }) => a > b ? 1 : -1)
```
To be correct in all cases the comparator needs to return `0` when `a` and `b` are equivalent. 

Sorting text strings that may contain marks benefits from using `localeCompare` to correctly compare strings; `localeCompare` correctly returns `0` when the strings are equivalent.
```
const comparator = ({ value: a}, { value: b }) => a.localeCompare(b);
```

packages/11ty/_includes/components/object-cards-list.webc
```diff
@@ -1,22 +1,46 @@
...
-         image-path="${imagePath}"
-         object-name="${objectName}"
-         object-name-html="${objectNameHtml}"
-         url="${url}"
-         ${this.webc.renderAttributes(objectData)}
-       ></object-card>
-     </li>`
-  }).join('')
+<ol class="object-cards">
+  <li webc:for="object of objects" :data-object-id="object.id" @html="getObjectCard(object)"></li>
```
mphstudios, line 2: Avoiding using a template and JavaScript rendered markup in that template may make these components easier to read, something like the following is likely easier to understand:
```suggestion
  <li webc:for="object of objects" :data-object-id="object.id">
    <object-card :url="object.url" $renderAttributes(object.objectData)></object-card>
  </li>
```

packages/11ty/_includes/components/object-card.webc
```diff
@@ -1,48 +1,61 @@
...
+              <use xlink:href="#arrow-forward-icon"></use>
+            </switch>
+          </svg>
+        </span>
       </span>
-    </span>
-    <!--- @TODO style fields as inline-flex --->
-    <!--- ${renderFields()} --->
+      <span webc:else class="object-card__field" @html="attribute.value"></span>
+    </template>
```
mphstudios, line 18: I think that this can be rendered without the conditional by preparing the data in the setup script to fit a dedicated field in the card or a `slot` and/or using a CSS first element selector. The svg icon might also be positioned using CSS flexbox or grid rather than in the `span` inline with the value.

For the display of label: value data a HTML DefinitionList may be more semantic, for example:
```
<h2>{{ objectCard.title || objectCard.name }}</h2>
<dl>
  <dt>artist</dt>
  <dd>unknown</dd>
  <dt>date</dt>
  <dd>circa 1600</dd>
  <dt>medium</dt>
  <dd>metal, wood (ebony)</dd>
</dl>
```
Which more easily allows the CSS to capitalize the first letter of the definition term `dt` element or add styles.

packages/11ty/_includes/components/object-filters.webc
anderspollack: this works with `localeCompare`, we just can't destructure + rename `a`/`b` from `{ value: a, value: b }`

packages/11ty/_includes/components/object-filters/object-cards-list.webc
```diff
@@ -1,22 +1,9 @@
...
-         object-name-html="${objectNameHtml}"
-         url="${url}"
-         ${this.webc.renderAttributes(objectData)}
-       ></object-card>
-     </li>`
-  }).join('')
-};
+<ol class="object-cards">
+  <li webc:for="object of objects" :data-object-id="object.id">
+    <object-card :url="object.url" :data="object.objectData"></object-card>
```
mphstudios: We might prefix the attributes here with `@` to use them as component properties so that they are excluded from the output.
```suggestion
    <object-card :@url="object.url" :@data="object.objectData"></object-card>
```

mphstudios (approved):
@anderspollack this pull-request has come together well, let's merge this feature.

PR #851, opened 2023-10-03
Add support for objects with `field` and `label` keys in `objects.yaml` `object_filters` field to specify Filter label names

anderspollack:
Filter field labels are humanized object keys by default, but now users can specify custom field labels in `objects.yaml`:
```yaml
object_filters:
  - field: type
    label: Object Type
  - theme
  - material
```
in this case, the label for the `type` field will render as **"Object Type"**

### DIFF
--- a/packages/11ty/_includes/components/object-filters.webc
+++ b/packages/11ty/_includes/components/object-filters.webc
@@ -1,9 +1,6 @@
 <script webc:type="js">
 const { filters } = this.webc.attributes;
-/**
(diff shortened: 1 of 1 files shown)


geealbers:
@anderspollack, the custom labels are being applied but the filters aren't filtering. You can see it in the URL when you apply a filter. For example, filtering for type "Container" results in localhost:8080/things/?=Container instead of localhost:8080/things/?type=Container.

PR #856, opened 2023-10-09
DEV-16151: Add prev/next navigation buttons to `_layouts/objects-page.webc`

anderspollack:
This update renders publication navigation buttons below the objects catalog in `objects-page` layout.
Following this same pattern, we are now calling the `pageHeader` shortcode from the markup `@html` attribute, which allows us to remove the `page-header.webc` wrapper component


mphstudios (approved):
The separate `webc` components for the `page-header` and `page-buttons` was nice though perhaps not as they were simply wrappers for the JavaScript functions.

PR #858, opened 2023-10-10
DEV-16233: Ensure `:href` is a dynamic attribute in `object-card.webc`

anderspollack:
Not sure when this regressed but links should point to object pages now, not literally `"this.webc.attributes.url"`

### DIFF
--- a/packages/11ty/_includes/components/object-filters/object-card/object-card.webc
+++ b/packages/11ty/_includes/components/object-filters/object-card/object-card.webc
@@ -1,4 +1,4 @@
-<a class="object-card" href="this.webc.attributes.url">
+<a class="object-card" :href="this.webc.attributes.url">
   <div class="object-card__content">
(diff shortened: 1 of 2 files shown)


anderspollack:
@geealbers I think this should also fix DEV-16234 as well, these were typos. Currently, you can set an `alt` on a thumbnail if you provide the thumbnail as an object like this:
```yaml
object:
  - thumbnail:
    src: 'path/to/image`
    alt: 'Image Alt'
```

anderspollack:
@geealbers I fixed a separate bug in the `alt` function and ensured those are being markdownified, but the error you are seeing is because of the custom markup in the object `armchair.md` file:

```liquid
{% for item in object %}
<img src="/_assets/images/{{ item.thumbnail }}" width="150"/>
```

is still looking for `item.thumbnail`, but it needs `item.thumbnail.src` in this instance

geealbers:
Doh! of course, sorry about that @anderspollack. thanks

geealbers (approved):

PR #862, opened 2023-10-10
DEV-16154: EPUB/PDF styles for object filters

anderspollack:
See https://github.com/thegetty/quire-starter-default/pull/40 to add a few extra styles to give object cards a nice off-white background in PDF output

Screenshot 2023-10-10 at 5 48 21 PM


mphstudios (dismissed):
@anderspollack this looks really good.

PR #864, opened 2023-10-11
DEV-16148: Add support for `object_filters: false` to disable filtering and hide UI

anderspollack:
set `object_filters: false` either in `things/index.md` front matter or `objects.yaml` to disable filter UI. When `object_filters` is not specified in either place, filters for _all_ object properties will still render as before.

### DIFF
--- a/packages/11ty/_includes/components/object-filters/object-filters.webc
+++ b/packages/11ty/_includes/components/object-filters/object-filters.webc
@@ -1,5 +1,5 @@
 <div class="object-filters__list">
-  <div webc:for="filter of getFilters(this.$data, objects)" class="object-filters__item">
+  <div webc:for="filter of getFilters(getFilterOrder($data), objects)" class="object-filters__item">
     <label :for="filter.id" @html="filter.fieldLabel"></label>
     <div class="object-filters__select">
       <select :id="filter.id" data-control-filter :data-control-name="filter.fieldName">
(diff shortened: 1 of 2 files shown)


packages/11ty/_includes/components/object-filters/object-filters.webc
```diff
@@ -79,9 +81,26 @@ const getUniqueOptionsFromObjectData = (objects, fieldName) =>
...
+      /*
+       * When $data.object_filters is explicitly set to `false`, use it instead of
+       * global $data.objects.object_filters
+       */
+    case $data.object_filters === false:
+      return false;
+    default:
+      return $data.objects.object_filters;
+  }
+};
```
mphstudios, line 97:
```suggestion
/**
 * When object_filters is set to `false` filters will not be rendered
 */
const getFilterOrder = ($data) => {
  return $data.object_filters !== false
    ? $data.object_filters || $data.objects.object_filters;
    : false;
  }
};
```

packages/11ty/_includes/components/object-filters/objects-catalog.webc
```diff
@@ -7,7 +7,7 @@
       <object-filters :@objects="objects"></object-filters>
     </div>
     <div class="collapsible__controls">
-      <div class="object-filters__controls">
+      <div webc:if="!areFiltersDisabled($data)" class="object-filters__controls">
```
mphstudios, line 10: Using a positive assertion here may be easier to read
```suggestion
      <div webc:if="displayFilters($data)" class="object-filters__controls">
```

packages/11ty/_includes/components/object-filters/objects-catalog.webc
```diff
@@ -80,6 +80,10 @@ const sortByOrderKey = (a, b) => {
   return a.order > b.order ? 1 : -1;
 };
 objects = objects.sort(sortByOrderKey);
+
+const areFiltersDisabled = ($data) => {
+  return $data.object_filters === false || $data.objects.object_filters === false;
+}
```
mphstudios, line 86:
```suggestion
const hasFilters = ($data) => {
  return $data.object_filters !== false || $data.objects.object_filters !== false;
}
```

mphstudios (changes requested):
See inline comments.

anderspollack:
@mph changes in comments have been applied + tested

mphstudios (approved):

PR #865, opened 2023-10-11
DEV-15887: Include title in epub manifest if no subtitle

anderspollack:
This `epubTitle` function should now work in all permutations of `readingLine` and `subtitle` being defined (assuming title is defined, which if it isn't causes problems elsewhere)

### DIFF
--- a/packages/11ty/_plugins/transforms/outputs/epub/manifest.js
+++ b/packages/11ty/_plugins/transforms/outputs/epub/manifest.js
@@ -85,10 +85,16 @@ module.exports = (eleventyConfig) => {
    * Publication title, subtitle, and reading line
    */
   const pubTitle = () => {
-    if (subtitle && readingLine) {
-      return `${title}: ${subtitle} ${readingLine}`
-    } else if (subtitle) {
-      return `${title}: ${subtitle}`
+    const joiner = title.endsWith('?') || title.endsWith('!') ? '' : ':';
+    switch (true) {
+    case subtitle && readingLine:
+      return `${title}${joiner} ${subtitle} ${readingLine}`
+    case readingLine:
+      return `${title} (${readingLine})`
+    case subtitle:
+      return `${title}${joiner} ${subtitle}`
+    default:
+      return title
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/11ty/_plugins/transforms/outputs/epub/manifest.js
```diff
@@ -85,10 +85,16 @@ module.exports = (eleventyConfig) => {
    * Publication title, subtitle, and reading line
    */
   const pubTitle = () => {
-    if (subtitle && readingLine) {
-      return `${title}: ${subtitle} ${readingLine}`
-    } else if (subtitle) {
-      return `${title}: ${subtitle}`
+    const joiner = title.endsWith('?') || title.endsWith('!') ? '' : ':';
```
mphstudios, line 88: More robustly, if the title does not end with punctuation, use a separator.
```suggestion
    const separator = title.match(/[.,:!?]$/ '' : ':';
```

packages/11ty/_plugins/transforms/outputs/epub/manifest.js
```diff
@@ -85,10 +85,16 @@ module.exports = (eleventyConfig) => {
...
+    switch (true) {
+    case subtitle && readingLine:
+      return `${title}${joiner} ${subtitle} ${readingLine}`
+    case readingLine:
+      return `${title} (${readingLine})`
+    case subtitle:
+      return `${title}${joiner} ${subtitle}`
+    default:
+      return title
```
mphstudios, line 97:
```suggestion
      case subtitle && readingLine:
        return `${title}${separator} ${subtitle} ${readingLine}`
      case readingLine:
        return `${title} (${readingLine})`
      case subtitle:
        return `${title}${separator} ${subtitle}`
      default:
        return title
```

mphstudios (changes requested):
We can proactively test for all punctuation when determining how to separate the `title` `subtitle` string.

packages/11ty/_plugins/transforms/outputs/epub/manifest.js
mphstudios: An alternate functional approach
```
function endsWithPunctuation (string) {
  const punctuation = '.,:!?';
  return punctuation.some((character) => string.endsWith(character));
}
```

anderspollack:
@mphstudios changes applied + tested

mphstudios (approved):

PR #866, opened 2023-10-12
DEV-15886: Ensure relative links with hashes are properly transformed for EPUB output

anderspollack:


anderspollack:
@mphstudios this works, but it is not the most elegant solution; thinking about how to model these relative link `href`s with/without hashes as a `new relativeURL() => { hash: '...', pathname: '...', href: '..' }`

packages/11ty/_plugins/transforms/outputs/epub/transform.js
```diff
@@ -143,12 +143,27 @@ module.exports = function(eleventyConfig, collections, content) {
...
+
+      return {
+        hash,
+        href,
+        pathname
+      }
+    }
+
+    const { hash, pathname } = relativeURL(href)
+
```
mphstudios, line 160: We could use a `URL` instance here with `try` `catch` `finally` blocks...
```suggestion
    function relativeUrl (path) {
      const base = eleventyConfig.baseURL || 'http://localhost'
      let url;
      try {
        url = new URL(path)
      } catch (TypeError) {
        url = new URL(path, base)
      } finally {
        return url
      }
    }
    const { hash, pathname } = relativeUrl(href)

```

mphstudios (approved):

PR #885, opened 2023-12-05
DEV-16534: WebC Figure image refactor

anderspollack:
This adds:
- `_components/quire-figure.webc`: defines a `<figure>` with an `id` and conditionally renders `figureAudio`, `figureTable`, `figureVideo`, and `figureImageWebc`
- `_components/figure-image.webc`: conditionally renders `<canvas-panel>`, `<image-sequence>`, `<image-service>`, and `<image-tag>` components based on `mediaType`. Registered as an eleventy shortcode named `figureImageWebc`
- `_components/image-tag.webc`: renders a `` tag with alt text

**Note:** Media embed links in video/audio figures are currently rendering in web output; these links should only show in print output. Image Sequences still need to be addressed as well -- see https://jira.getty.edu/browse/DEV-16542

@mphstudios I was originally planning on doing these figure image component webc refactors in isolation, but refactoring the `_plugins/shortcodes/figure.js` to render `quire-figure.webc` ended up being much simpler -- we can still use the JS components for video, audio, and table figures via `@html="..."`. Interestingly enough, none of these actually need the `quire-data` JSON since we can just use our `getFigure(id)` filter -- I anticipate that will still be necessary / useful for refactoring the lightbox/modal.


packages/11ty/.eleventy.js
```diff
@@ -171,7 +171,7 @@ module.exports = function(eleventyConfig) {
    * @property {Boolean} useTransform - Use WebC transform to process all HTML output
    */
   eleventyConfig.addPlugin(pluginWebc, {
-    components: '_includes/components/**/*.webc',
+    components: ['_includes/components/**/*.webc', '_components/**/*.webc'],
```
mphstudios, line 174: Listing the default, which is `'_components/**/*.webc'`, first.
```suggestion
    components: ['_components/**/*.webc', '_includes/components/**/*.webc'],
```

packages/11ty/_components/figure-image.webc
```diff
@@ -0,0 +1,61 @@
+<template webc:nokeep>
+  <canvas-panel webc:if="isCanvas(id)" @attributes="attributes(id, image_dir).canvasPanel"></canvas-panel>
+  <image-service webc:elseif="isImageService(id)" @attributes="attributes(id, image_dir).imageService"></image-service>
+  <image-sequence webc:elseif="isSequence(id)" @attributes="attributes(id, image_dir).imageSequence"></image-sequence>
+  <image-tag webc:else @attributes="attributes(id, image_dir).imageTag"></image-tag>
+</template>
+<script webc:setup>
```
mphstudios, line 7:
```suggestion

<script webc:setup>
```

packages/11ty/_components/figure-image.webc
```diff
@@ -0,0 +1,61 @@
+<template webc:nokeep>
+  <canvas-panel webc:if="isCanvas(id)" @attributes="attributes(id, image_dir).canvasPanel"></canvas-panel>
+  <image-service webc:elseif="isImageService(id)" @attributes="attributes(id, image_dir).imageService"></image-service>
+  <image-sequence webc:elseif="isSequence(id)" @attributes="attributes(id, image_dir).imageSequence"></image-sequence>
+  <image-tag webc:else @attributes="attributes(id, image_dir).imageTag"></image-tag>
+</template>
```
mphstudios, line 6: Could the template be written using a dynamic webc component, `webc:is="component(id)"`, rather than a webc conditional statements?

packages/11ty/_components/figure-image.webc
```diff
@@ -0,0 +1,61 @@
...
+    margin,
+    preset,
+    region,
+    src,
+    startCanvas,
+    textEnabled,
+    textSelectionEnabled,
+    virtualSizes,
+    width
+  } = this.getFigure(id)
```
mphstudios, line 25: Destructuring *all* of the `figure` properties may make this less readable.
```suggestion
  const figure = this.getFigure(id)
```

packages/11ty/_components/figure-image.webc
mphstudios: Apparently, `webc:is` does not support dynamic values; see https://github.com/11ty/webc/issues/143

mphstudios: We might add a comment on the implementation quirks, something like
```
<!--- 
  Nota bene: `webc:is` does not support dynamic values (see 11ty/webc#143);
  use webc conditional statements and call the `attributes` function multiple times rather than using a fancier pattern of implementation.
--->
```

mphstudios (approved):

PR #927, opened 2024-04-30
updates default sort

bulbil:

*For the Title of this pull request, please use the format "Type/Issue-#: Brief description." For Type, the options are Fix, Feature, Docs, or Chore. Issue-# is only needed if this pull request addresses an [existing issue](https://github.com/thegetty/quire/issues).*

**Before submitting your PR, make sure that you can run the following commands: `quire preview`, `quire build`, `quire pdf`, and `quire epub`. If any of those core quire commands throw errors/yield unexpected behavior, we will NOT review the PR!**

### Checklist 

- [x] I have read the [CONTRIBUTING.md](https://github.com/thegetty/quire/blob/main/CONTRIBUTING.md) file

- [x] I have made my changes in a new branch and not directly in the main branch

Developing locally I found that when `order` is not set or has many duplicate values, default sort was not [stable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description).

I think Hugo's default sort fallbacks and the way it uses `weight` is significantly different from the current behavior. I created a `useCustomSort` config key to implement a more flexible TOC sort based on title string.

At the very least, I propose you `getAllSorted` first in order to ensure that order is not different between builds when `order` is missing.

### DIFF
--- a/packages/11ty/_plugins/collections/index.js
+++ b/packages/11ty/_plugins/collections/index.js
@@ -1,9 +1,9 @@
-const filters = require('./filters')
-const sortCollection = require('./sort')
+const filters = require("./filters");
+const sortCollection = require("./sort");
 
 /**
  * Add Collections and Apply Transforms
- * 
+ *
  * Nota bene: The Eleventy API does not make collections data accessible
  * from the plugin context. Adding `collections` and `transforms` sequentially
  * in the same file allows access to `collections` data from `transforms`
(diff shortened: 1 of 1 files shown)


bulbil:
@anderspollack here's that super minor update I mentioned earlier today to do with sort -- not having been part of the conversations, not sure how helpful this is or what desired behavior is.

I just thought having order be inconsistent between builds wasn't ideal. I have a more elaborate custom sort but thought that would require more discussion.

As a sidenote, I have an idea about how to have `search-index.json` not include HTML from the parsed MD, be happy to chat about that. It's not the most elegant thing and it doesn't seem like a pressing concern -- but in case you are interested I'd be happy to write something up.

mphstudios (changes requested):
@bulbil having the collection sort by consistent across builds is more clear. I am aware that the linter may not be functioning properly though Quire uses code format rules `no-semi` and `single-quotes`, if you can update this pull-request we will review/test/merge the change. Thank you for the fix!

bulbil:
@mphstudios apologies, I realized I had updated linting in my Quire project but not in this fork.

mphstudios (approved):

PR #942, opened 2024-08-06
DEV-18240: Fix PDF config validation across old and new projects

cbutcosk:
This PR fixes DEV-18240, "`quire pdf` throws error in new version of CLI (rc.13)":
- Fixes dynamic import syntax to unwrap function property 
- Checks file presence of validator and moves import to file-exists scope


mphstudios (changes requested):
@cbutcosk thanks for fixing my fix; if you could bump the release candidate version we will merge this pull-request

mphstudios (approved):

PR #960, opened 2024-10-09
Refactor image sequences to paint a canvas element and lazily load images

cbutcosk:
This PR updates `q-image-sequence` to refine its loading and drawing behaviors:
- Draws images to a passed-by-reference `canvas` element
- Preloads 2 images images ahead of the current position and 2 images behind the current position in the sequence
- If configured `@interactive=false`, the component will only load the first image from the network

This PR has some issues before release:
- [x] ~Fetches that end up outside of the buffer due to user interaction should cancel~ -- refine fetch / paint behavior
- [x] Buffer a (scaled) percentage of sequence length before allowing interaction
- [x] Add %age of loaded buffer to loading indicator
- [x] Refine buffering status indicator
- [x] Refine preload / prebuffer behavior to ensure entire sequence is eventually loaded


packages/11ty/_includes/web-components/image-sequence/index.js
```diff
@@ -123,86 +132,125 @@ class ImageSequence extends LitElement {
...
-    const imageElements = this.images.map( (img,i) => {
-      fetch(img).then( this.imagesLoaded++ )
+  /**
+   * @function _fetchImage
+   * @param url {string} - image URL to fetch
+   * @param seqIndex {Number} - index to store this image 
+   * @param draw {Boolean} - whether to draw after fetching
+   * 
+   * Fetches `url`, converts it into a blob and stores the image data, optionally drawing to the canvas
+   **/
```
mphstudios, line 187:
```suggestion
   */
```

packages/11ty/_includes/web-components/image-sequence/index.js
```diff
@@ -123,86 +132,125 @@ class ImageSequence extends LitElement {
...
+      entries.forEach( (entry) => {
+        if (entry.isIntersecting) {
+          this.visible = true
+          observer.disconnect()
+        }
+      })
+    }, {root:null, threshold: 0.5})
+
+    io.observe(this)
+
```
mphstudios, line 226:
```suggestion
```

packages/11ty/_includes/web-components/image-sequence/index.js
```diff
@@ -280,56 +321,115 @@ class ImageSequence extends LitElement {
...
+   * Paints the `canvas` element with the image from this.index
+   **/
+  _paintCanvas(image) {
+    if (!this.canvasRef.value) {
+      return
+    }
+
+    if (!this.context) {
+      this.context = this.canvasRef.value.getContext('2d')
+    }
```
mphstudios, line 366: Use logical nullish assignment
```suggestion
    this.context ??= this.canvasRef.value.getContext('2d')
```

packages/11ty/_includes/web-components/image-sequence/index.js
```diff
@@ -123,86 +132,125 @@ class ImageSequence extends LitElement {
...
       state: true,
     },
+    intrinsicHeight: {
+      type: Number,
+      state: true,
+    },
+    intrinsicWidth: {
+      type: Number,
+      state: true,
+    },
```
mphstudios, line 160: sort properties
```suggestion
    images: {
      type: Array,
      state: true,
    },
    imageUrls: {
      attribute: 'image-urls',
      type: String,
    },
    index: {
      type: Number,
    },
    intrinsicHeight: {
      type: Number,
      state: true,
    },
    intrinsicWidth: {
      type: Number,
      state: true,
    },
```

mphstudios (commented):
@cbutcosk what are your thougts about replacing the underscore convention with [private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties), for example `#privateFoo` instead of `_privateFoo`?

mphstudios (commented):
It may help make this Lit component more readable to move the css to a separate module in the component directory, for example:

`packages/11ty/_includes/web-components/image-sequence/styles.js`
```
import { css } from 'lit'
export default css`...`
```
which can we then import in the `image-sequence` module
```
import { LitElement } from 'lit'
import styles from './styles.js'

class ImageSequence extends LitElement {
  static styles
}
```

packages/11ty/_includes/web-components/image-sequence/index.js
```diff
@@ -1,208 +1,129 @@
...
+        if (entry.isIntersecting) {
+          this.visible = true
+          observer.disconnect()
+        }
+      })
+    }, {root:null, threshold: 0.5})
+
+    io.observe(this)
+
   }
```
mphstudios, line 100: linting example
```suggestion
  connectedCallback() {
    super.connectedCallback()

    // Observes this component against the viewport to trigger image preloading
    const io = new IntersectionObserver((entries, observer) => {
      entries.forEach((entry) => {
        if (entry.isIntersecting) {
          this.visible = true
          observer.disconnect()
        }
      })
    }, { root:null, threshold: 0.5 })

    io.observe(this)
  }
```

packages/11ty/_includes/web-components/image-sequence/index.js
```diff
@@ -1,208 +1,129 @@
...
+
+        if (draw) {
+          this.#paintCanvas(bmp)        
+        }
+
+      })
+      .catch( (err) => {
+        console.error(err)
+      })
+  }
```
mphstudios, line 83: linting suggestions...
```suggestion
  /**
   * Fetches `url`, converts it into a blob and stores the image data, optionally drawing to the canvas
   * @function #fetchImage
   * @private
   * @param {String} url - image URL to fetch
   * @param {Number} seqIndex - index to store this image 
   * @param {Boolean} draw - whether to draw after fetching
   */
  #fetchImage(url, seqIndex, draw) {
    const request = new Request(url)
    
    fetch(request)
      .then((response) => response.blob())
      .then((blob) => window.createImageBitmap(blob))
      .then((bmp) => {
        if (this.intrinsicHeight === 0) {
          this.intrinsicHeight = bmp.height
          this.intrinsicWidth = bmp.width   
        }
        this.images[seqIndex] = bmp
        if (draw) {
          this.#paintCanvas(bmp)
        }
      })
      .catch((error) => {
        console.error(error)
      })
  }
```

cbutcosk: @mphstudios I changed this code substantially-enough since this lint tweak that the github tool won't say so, but I *believe* these lint updates are all incorporated into the branch now. `npm run lint` passes on this branch and separately I've issued #965 which quiets the other linting noise that was masking the lint errors in this branch.

cbutcosk: Erm.. Or, nevermind I see what I missed in these comments -- added and resolved

mphstudios (approved):

PR #973, opened 2024-12-02
Fix `DEV-19079`: `ref` tags that point to uninitialized canvas-panel components

cbutcosk:
This PR updates the client-side handling code for `ref` components to:
- Eagerly add lightbox preload attributes to the ref'd slide
- Apply a delay to canvas-panel region transitions to ensure underlying network resources have loaded
- Fixed an issue where video and table `ref` tags did not function (`DEV-18548`)
- Fixed an issue where image-sequence `@rotate-to-index` attributes would under-run the buffer toward the end of the animation

### DIFF
--- a/packages/11ty/content/_assets/javascript/application/canvas-panel.js
+++ b/packages/11ty/content/_assets/javascript/application/canvas-panel.js
@@ -93,6 +93,9 @@ const goToFigureState = function ({
   })
 
   if (figureSlide) {
+    // Set the preloading attribute to ensure the slide has a chance to load
+    figureSlide.dataset.lightboxPreload = true
+
     const lightbox = figureSlide.closest('q-lightbox')
     lightbox.currentId = figureId
   }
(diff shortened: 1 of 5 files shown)


packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -299,14 +302,25 @@ const update = (id, data) => {
...
-            duration: 2000
+
+      setTimeout( () => { 
+        element.transition(tm => {
+          // If the transitionManager is not present, wait to load and manually set a region
+          if (!tm) {
+            setTimeout( () => {
+              element.setAttribute('region',region)          
+            },500)
+            return
```
mphstudios, line 313:
```suggestion
      setTimeout(() => { 
        element.transition((tm) => {
          // If the transitionManager is not present, wait to load and manually set a region
          if (!tm) {
            setTimeout(() => element.setAttribute('region', region), 500)
            return
```

packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -299,14 +302,25 @@ const update = (id, data) => {
...
-        })
-      })
+          tm.goToRegion(target, {
+              transition: {
+                easing: element.easingFunctions().easeOutExpo,
+                duration: 2000
+              }
+          })
+        }) 
+      },300)
```
mphstudios, line 322:
```suggestion
      }, 300)
```

packages/11ty/content/_assets/javascript/application/canvas-panel.js
```diff
@@ -299,14 +302,25 @@ const update = (id, data) => {
...
+
```
mphstudios, line 323:
```suggestion
          tm.goToRegion(target, {
            transition: {
              easing: element.easingFunctions().easeOutExpo,
              duration: 2000
            }
          })
        }) 
      }, 300)
```

mphstudios (changes requested):
see inline linting suggestions

mphstudios (approved):

PR #997, opened 2025-02-19
Fix `DEV-19745`: quire CLI fails to build 11ty v2-based projects 

cbutcosk:
This PR updates the CLI wrapper that the quire CLI uses to call into eleventy to parse the major version of the 11ty version it's wrapping. Eleventy v3 now using ES modules means that its extensions determine whether core-js or esm is used so the command extension called needs to change accordingly:
- For v3 projects, use `node_modules/@11ty/eleventy/cmd.cjs`
- For v2 projects, continue to use `node_modules/@11ty/eleventy/cmd.js`

### DIFF
--- a/packages/cli/src/lib/11ty/cli.js
+++ b/packages/cli/src/lib/11ty/cli.js
@@ -25,7 +25,8 @@ const factory = (options = {}) => {
   const eleventyModuleDir = path.join(eleventyRoot, 'node_modules', '@11ty', 'eleventy')
   const packagePath = path.join(eleventyModuleDir,'package.json')  
 
-  const pack = fs.readFileSync(packagePath)
+  const pack = JSON.parse(fs.readFileSync(packagePath))
+
   const { version } = pack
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/11ty/cli.js
```diff
@@ -17,9 +18,24 @@ const factory = (options = {}) => {
...
+
+  const { version } = pack
+
+  const eleventyMajorVer = version.split('.').at(0)
+
+  let eleventy = path.join(eleventyModuleDir,'cmd.cjs')
+
+  if ( Number(eleventyMajorVer) < 3 ) {
+    eleventy = path.join(eleventyModuleDir,'cmd.js')
+  }
```
mphstudios, line 38:
```suggestion
  if (Number(eleventyMajorVer) < 3) {
    eleventy = path.join(eleventyModuleDir,'cmd.js')
  }
```

mphstudios (approved):

Issue #989, opened 2025-01-21
Publication search JSON is not serialized to configured output directory


cbutcosk:
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [x] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

Quire's integration around 11ty is supposed to allow users to change the `ELEVENTY_OUTPUT` env var to direct site build outputs to a directory. This works mostly as expected but.. 

### Actual Behavior

the search plugin serializes its json data to `_site/search.json` in all cases. 

### Steps to Reproduce

- Create a new quire publication and change to the publication directory
- Build the publication (`ELEVENTY_OUTPUT="_other_site" quire build`)
- Check _other_site/search.json

### Version Numbers

1.0.0-rc17 but other versions likely affected

### Relevant Terminal/Shell Output

```shell

```

PR #1001, opened 2025-02-26
Fix - ensure search plugin writes JSON to configured output directory

cbutcosk:
This PR fixes an issue where the search plugin does not correctly use a build run's configured publicDir and outputDir by  replacing hard coded string values with the appropriate entries from directoryConfig.


cbutcosk:
Resolves #989

mphstudios (approved):

PR #1015, opened 2025-04-07
Resolve import URLs to paths for mutation and back to URLs for import

cbutcosk:
This PR fully resolves DEV-19999 by using node's internal URL -> File Path and File Path -> URL transforms for doing path math around `renderOutputs`. Primarily this affects figures but also table of contents items and copyright icons. Also resolves DEV-20261

### DIFF
--- a/packages/11ty/_includes/components/figure/audio/index.js
+++ b/packages/11ty/_includes/components/figure/audio/index.js
@@ -1,8 +1,9 @@
+import path from 'node:path'
 import { fileURLToPath } from 'node:url'
 
 export default function (eleventyConfig) {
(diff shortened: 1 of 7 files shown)


mphstudios (approved):

mphstudios (commented):
💭 this pattern might become a `resolveModulePath` function
```
import { dirname, resolve } from 'node:path'
export resolveModulePath = (url) => resolve(dirname(fileURLToPath(url)))
```
so that elsewhere we can write the following
```
import resolveModulePath from '#lib/resolveModulePath`
resolveModulePath(import.meta.url)
```
however that does not meet all of our usecases, however the [`unjs/exsolve`](https://github.com/unjs/exsolve) package does have methods to address these use cases.

PR #1025, opened 2025-04-16
Improve asset path safety on Windows

cbutcosk:
This PR refines the image source paths in HTML and the epub asset handling to ensure that both are safe on windows. In particular it:
- Forces img@src attributes to use forward slash separators to ensure `vite`'s resolve process (or quire's vite config?) can find assets on windows
- Refines asset path checks in image tags, image print representations, and object pages representations. This resolves _many_ latent path bugs on Windows
- Normalizes paths emitted in epub transform helper functions to use platform-specific versions for asset moves (eg, copying assets to the epub bundle)

### DIFF
--- a/packages/11ty/_includes/components/figure/image/image-tag.js
+++ b/packages/11ty/_includes/components/figure/image/image-tag.js
@@ -15,7 +15,13 @@ export default function (eleventyConfig) {
   const { imageDir } = eleventyConfig.globalData.config.figures
 
   return function ({ alt = '', src = '', isStatic = false, lazyLoading = 'lazy' }) {
-    const imageSrc = src.startsWith('http') || isStatic ? src : path.join(imageDir, src)
+    const extOrIiifRegex = new RegExp(/^(https?:\/\/|\/iiif\/|\\iiif\\)/)
+    let imageSrc = extOrIiifRegex.test(src) || isStatic ? src : path.join(imageDir, src)
+
+    // HACK: If an URL-unsafe path separator has made it this far, remove it
+    if (path.sep !== '/') {
+      imageSrc = imageSrc.replaceAll(path.sep,'/')
+    }
 
     return html`
       <img
--- a/packages/11ty/_plugins/transforms/outputs/epub/manifest.js
+++ b/packages/11ty/_plugins/transforms/outputs/epub/manifest.js
@@ -61,7 +61,11 @@ export default (eleventyConfig) => {
       return
     }
 
-    return path.join(imageDir.replace(/^\//, ''), image)
+    // Remove leading absolute pathing
+    let imageDirPath = path.normalize(imageDir)
+    if (imageDirPath.split(path.sep).at(0) === '') imageDirPath = imageDirPath.split(path.sep).slice(1).join(path.sep)
+    
+    return path.join(imageDirPath, image)
   }
 
   /**
--- a/packages/11ty/_plugins/transforms/outputs/epub/transform.js
+++ b/packages/11ty/_plugins/transforms/outputs/epub/transform.js
@@ -21,7 +21,7 @@ export default function (eleventyConfig, collections, content) {
   const write = writer(outputDir)
 
   /**
-   * Gather asset filepaths
+   * Gather asset filepaths, normalizing to platform paths for consumers
    *
    * @param      {HTMLElement}  element
    */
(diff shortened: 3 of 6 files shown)


packages/11ty/_includes/components/figure/image/image-tag.js
```diff
@@ -15,7 +15,13 @@ export default function (eleventyConfig) {
   const { imageDir } = eleventyConfig.globalData.config.figures
 
   return function ({ alt = '', src = '', isStatic = false, lazyLoading = 'lazy' }) {
-    const imageSrc = src.startsWith('http') || isStatic ? src : path.join(imageDir, src)
+    const extOrIiifRegex = new RegExp(/^(https?:\/\/|\/iiif\/|\\iiif\\)/)
+    let imageSrc = extOrIiifRegex.test(src) || isStatic ? src : path.join(imageDir, src)
+
+    // HACK: If an URL-unsafe path separator has made it this far, remove it
+    if (path.sep !== '/') {
+      imageSrc = imageSrc.replaceAll(path.sep,'/')
```
mphstudios, line 23:
```suggestion
      imageSrc = imageSrc.replaceAll(path.sep, '/')
```

packages/11ty/_plugins/transforms/outputs/epub/manifest.js
```diff
@@ -61,7 +61,11 @@ export default (eleventyConfig) => {
       return
     }
 
-    return path.join(imageDir.replace(/^\//, ''), image)
+    // Remove leading absolute pathing
+    let imageDirPath = path.normalize(imageDir)
+    if (imageDirPath.split(path.sep).at(0) === '') imageDirPath = imageDirPath.split(path.sep).slice(1).join(path.sep)
```
mphstudios, line 66: 💭 this might be more clear?
```suggestion
     if (imageDirPath.startsWith('/')) {
       imageDirPath = imageDirPath.slice(1)
     }
```

packages/11ty/_plugins/transforms/outputs/epub/transform.js
```diff
@@ -30,12 +30,10 @@ export default function (eleventyConfig, collections, content) {
...
       if (!src) return
-      const pattern = `^(${imageDir}|/${iiifOutputDir})`
-      const regex = new RegExp(pattern, 'g')
-      if (src.match(regex)) {
-        const relativePath = src.replace(/^\//, '')
-        assets.push(relativePath)
-      }
+
+      const relativePath = path.normalize(src).split(path.sep).at(0) === '' ?
+                              path.normalize(src).split(path.sep).slice(1).join(path.sep) : src
```
mphstudios, line 35: 💭 seeing this pattern more than once in the code, removing a leading path separator in order to ensure the path is relative, might be an indication of a change earlier in the data manipulation that needs to be made? That being possible we might also abstract that into a method like above, something like following is a bit more clear
```
const preparePath(pathArg) => {
  let pathString = path.normalize(pathArg)
  return (pathString.startsWith(path.sep)) ? pathString.slice(1) : pathString
}
```

packages/11ty/_plugins/transforms/outputs/epub/manifest.js
cbutcosk: @mphstudios The bracing and removal of complexity is fine / welcome but the check needs to be platform-specific ala:
```
if (imageDirPath.startsWith(path.sep)) { imageDirPath = imageDirPath.slice(1) }
```

packages/11ty/_plugins/transforms/outputs/epub/transform.js
cbutcosk: @mphstudios Oh yes quite, and probably both -- the transforms should share a function like this (pdf and epub xforms are mostly about re-rooting [src] attribs) and the figures plugin should pre-compose asset paths so componentry isn't having to figure out pathing (and thus catch bad / unnormalized-to-this-point paths as they're harvested out of HTML)

mphstudios (approved):

mphstudios (dismissed):
@cbutcosk if you add a `CHANGELOG` to this pull request it looks good to merge.

cbutcosk:
@mphstudios Done! Ready to merge if you re-review

mphstudios (approved):

PR #1033, opened 2025-05-23
Fix failing Windows tests -- join IIIF resource IDs with '/' instead of path.sep

cbutcosk:
This PR corrects URIs in IIIF manifests, which had been failing on all platforms and had additional platform-specific failures for Windows.

### DIFF
--- a/packages/11ty/_plugins/figures/annotation/index.js
+++ b/packages/11ty/_plugins/figures/annotation/index.js
@@ -63,10 +63,11 @@ export default class Annotation {
       (annotationCount === 0 || src === figureSrc)
     const info = () => {
       if (!isImageService) return
-      const tilesPath = path.join(outputDir, name, tilesDirName)
-      const infoPath = path.join(tilesPath, 'info.json')
+      // NB: Not joining by path.sep because these are URLs
+      const tilesPath = [outputDir, name, tilesDirName].join('/')
+      const infoPath = [tilesPath, 'info.json'].join('/')
       try {
-        return new URL(path.join(baseURI, infoPath)).href
+        return new URL(`${baseURI}/${infoPath}`).href
       } catch (error) {
         logger.error(`Error creating info.json. Either the tile path (${tilesPath}) or base URI (${baseURI}) are invalid to form a fully qualified URI.`)
       }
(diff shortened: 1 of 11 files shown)


packages/11ty/_lib/urlPathJoin/index.js
```diff
@@ -0,0 +1,23 @@
...
+ **/
+
+const urlPathJoin = (url, ...segs) => {
+  const base = new URL(url)
+  const basePath = base.pathname
+
+  base.pathname = path.join(basePath, ...segs)
+
+  return base.href
+}
```
mphstudios, line 21: ⚙️ I think the following is a more idiomatic use of the node api
```suggestion
const urlPathJoin = (baseUrl, ...segments) => {
  const url = new URL(path.join(...segments), baseUrl)
  return url.toString()
}
```

packages/11ty/_lib/urlPathJoin/index.js
```diff
@@ -0,0 +1,23 @@
+import { posix as path } from 'node:path'
+/**
+ * @function urlPathJoin
+ *
+ * @argument {String} url
+ * @argument {Iterable} segs
```
mphstudios, line 6:
```suggestion
 * @argument {String} baseUrl
 * @argument {Iterable} segments
```

packages/11ty/_lib/urlPathJoin/index.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Parses `url` and safely joins all of `segs` to its pathname
+ *
+ * @return {String}
+ *
+ **/
```
mphstudios, line 12:
```suggestion
 * @return {String}  URL string
 */
```

mphstudios (changes requested):
@cbutcosk see inline comments

packages/11ty/_lib/urlPathJoin/index.js
```diff
@@ -0,0 +1,23 @@
...
+const urlPathJoin = (url, ...segs) => {
+  const base = new URL(url)
+  const basePath = base.pathname
+
+  base.pathname = path.join(basePath, ...segs)
+
+  return base.href
+}
+
+export default urlPathJoin
```
mphstudios, line 23: ⚙️ @cbutcosk the following may be a more idiomatic use of the node api
```suggestion
import path from 'node:path'

/**
 * @function urlPathJoin
 *
 * @argument {String} baseUrl
 * @argument {Iterable} segments
 *
 * Parses `url` and safely joins all of `segs` to its pathname
 *
 * @return {String}
 */
const urlPathJoin = (baseUrl, ...segments) => {
  const url = new URL(path.posix.join(...segments), baseUrl)
  return url.toString()
}

export default urlPathJoin
```

cbutcosk: @mphstudios The variable renames and shift away from renamed imports are 👍 (commits are ready to push) but relying on the relative URL rebasing of `new URL(input, base)` has some unpleasant corner cases for callers. For the pathname to join successfully `base` must end in '/' and `input` must not be an absolute path. Neither of these are the case for callers here, who in practice are expecting path.join()-style path joins, just with a base URL at the front (path.posix.join() normalizes the URL protocol's '//' to '/' in those cases).

PR #1038, opened 2025-06-05
Fix inclusion of SVG icons in PDF output

geealbers:
SVG icons were not being included properly in the PDF output due to a naming issue: Quire was creating `svgSymbolElements` in `_plugins/transforms/outputs/pdf/transform.js` but then calling for `svgElements` in `_layouts/pdf.liquid`. Rather than fixing the naming though, SVG icons can simply be added to the PDF layout directly, without need of the extra logic. This PR does that.

### DIFF
--- a/packages/11ty/_layouts/pdf.liquid
+++ b/packages/11ty/_layouts/pdf.liquid
@@ -17,10 +17,10 @@
   `pages` is an array of `section` elements from each PDF-having page in the publication.
   {% endcomment %}
 
+  {% icons %}
+  {% iconscc %}
+
   {% for page in pages %}
-    {% if forloop.first == true %}
-      {{ page.svgElements }}
-    {% endif %}
     {{ page.sectionElement }}
   {% endfor %}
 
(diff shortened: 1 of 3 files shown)


mphstudios (changes requested):
@geealbers the code being replaced by this pull requests was selectively including SVG elements rather than including all SVG icons, even if they are not used. Fixing the mismatch in the names is likely a better solution though I think we need to write tests for inclusion of the SVG elements.

geealbers:
@mphstudios I did some testing on this. The original code is also including all the icons, not just the ones that are being used on the page, because we're including all of the icons in `_layout/base.11ty.js` with:

```js
${this.icons(data)}
${this.iconscc(data)}
```

So, I can go back on this PR and use the original code in `_plugins/transforms/outputs/pdf/transform.js` and just fix the naming error in `_layouts/pdf.liquid`:

```
{% if forloop.first == true %}
  {{ page.svgSymbolElements }}
{% endif %}
```
But unless I'm missing something (which of course I could be) it seems the way I have the PR now is simpler, parallels what we're doing in `_layout/base.11ty.js`, and gives the same result.

```
{% icons %}
{% iconscc %}
```

PR #1057, opened 2025-08-24
Fix duplicate ID error in accordion shortcode (DEV-21555)

cbutcosk:
This PR addresses an issue (DEV-21555) where projects that use the accordion shortcode will not build without disabling ID checking in the configuration for the IdAttributePlugin in the project's `.eleventy.js`. The patch:
- Adds a test helper function that initializes the `Eleventy()` object and passes it on to later tests (helper in `packages/11ty/_tests/helpers/index.js`.
- Adds a test for duplicate IDs in the accordion shortcode when used with and without a footnote.
- Refactors `accordion` shortcode to remove duplicate markup and use `summary`/`details` tags in all outputs and sets details to open in PDF

### DIFF
--- a/packages/11ty/_plugins/globalData/index.js
+++ b/packages/11ty/_plugins/globalData/index.js
@@ -51,7 +51,12 @@ const checkForDuplicateIds = function (data, filename) {
  */
 export default function (eleventyConfig, directoryConfig) {
   const dir = path.resolve(directoryConfig.inputDir, '_data')
-  // console.debug(`[plugins:globalData] ${dir}`)
+
+  // Add directory config to globalData so that it is available to other plugins
+  eleventyConfig.addGlobalData('directoryConfig', directoryConfig)
+
+  if (!fs.existsSync(dir)) return eleventyConfig.globalData
+
   const files = fs.readdirSync(dir)
     .filter((file) => path.extname(file) !== '.md')
   const parse = parser(eleventyConfig)
@@ -74,9 +79,6 @@ export default function (eleventyConfig, directoryConfig) {
     checkForDuplicateIds(value, file)
     eleventyConfig.addGlobalData(key, value)
   }
-
-  // Add directory config to globalData so that it is available to other plugins
-  eleventyConfig.globalData.directoryConfig = directoryConfig
 
   return eleventyConfig.globalData
 }
--- a/packages/11ty/_tests/helpers/index.js
+++ b/packages/11ty/_tests/helpers/index.js
@@ -1,3 +1,5 @@
+import Eleventy from '@11ty/eleventy'
+
 /**
  * @function stubGlobalData
  *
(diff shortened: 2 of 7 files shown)


geealbers:
@cbutcosk I finished my QA on this work. Three things:

1. I put a PR ([#51](https://github.com/thegetty/quire-starter-default/pull/51)) into `quire-starter-default` that fixes styling issues. 
2. Even with that PR, there's one style/display issue I could figure out that's specific to paged.js detailed here: [DEV-21892](https://jira.getty.edu/browse/DEV-21892)
3. There were a couple issues with the EPUB, the main one being that even though `<details>` elements are allowed and work in EPUB, they also mostly don't work because not all EPUB readers will flow out more pages to accommodate the expanded content. The result is that you can get stuck on the single page without access to the content Simply setting the accordions to `open` and removing the `tabindex` should fix things up in EPUB though. Details here: [DEV-21893](https://jira.getty.edu/browse/DEV-21893)

Here's the testing page I threw together if it's useful: [essay-accordions.md](https://github.com/user-attachments/files/22123377/essay-accordions.md)

packages/11ty/_plugins/globalData/index.js
```diff
@@ -51,7 +51,12 @@ const checkForDuplicateIds = function (data, filename) {
  */
 export default function (eleventyConfig, directoryConfig) {
   const dir = path.resolve(directoryConfig.inputDir, '_data')
-  // console.debug(`[plugins:globalData] ${dir}`)
+
+  // Add directory config to globalData so that it is available to other plugins
+  eleventyConfig.addGlobalData('directoryConfig', directoryConfig)
+
+  if (!fs.existsSync(dir)) return eleventyConfig.globalData
+
```
mphstudios, line 59:
```suggestion
  if (fs.existsSync(dir)) {
    const files = fs.readdirSync(dir)
    const parse = parser(eleventyConfig)
    for (const file of files) {
      if (path.extname(file) === '.md') continue;
      const { name: key } = path.parse(file)
      const parsed = parse(path.join(dir, file))
      let value
      try {
        value = validateUserConfig(key, parsed)
      } catch (error) {
        logger.error(error)
        process.exit(1)
      }
      if (!key || !value) continue;
      checkForDuplicateIds(value, file)
      eleventyConfig.addGlobalData(key, value)
    }
  }
  
  return eleventyConfig.globalData
}
```

packages/11ty/_tests/helpers/index.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Initializes an Eleventy object suitable for rendering out shortcodes
+ *
+ **/
+const initEleventyEnvironment = async () => {
+  const elev = new Eleventy('../', '_site', { config: stubGlobalData })
+  await elev.init()
+
+  return elev
+}
```
mphstudios, line 38: ⚙️ prefer more rather than less explicit variable names
```suggestion
const initEleventyEnvironment = async () => {
  const eleventy = new Eleventy('../', '_site', { config: stubGlobalData })
  await eleventy.init()
  return eleventy
}
```

mphstudios (approved):

PR #1059, opened 2025-08-28
Resolve issue with figure asset paths in lightbox when publication URL has a pathname (DEV-21513)

cbutcosk:
This PR resolves an issue where lightbox assets would not load when the publication URL was set to something with a pathname (eg, `https://example.org/publication/`). An overview of the changes:
- Adds an integration test that builds a second publication at `test-publication-pathname` with the URL `http://localhost:8181/test-publication-pathname/` and persists it in CI to the browser testing stage.
- Creates a few helper functions to modify publication configuration and walk a publication sitemap.
- Fixes an issue where sitemap URLs did not include a pathname when present.
- Passes a `lightbox` argument to ImageElement and VideoElement render functions and prepends publication `pathname` when rendering for lightbox usage and pathname is not `/`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,6 +179,7 @@ jobs:
   # Tests the output of the publication build -- must be separate for the playwright executor
   browser_test:
     executor: pw-noble-development
+    # NB: taken from https://playwright.dev/docs/ci#sharding-in-circleci
     parallelism: 4
     steps:
       - attach_workspace:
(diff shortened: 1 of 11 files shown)


geealbers:
@cbutcosk I noticed you'd marked this as ready for review just when I was finishing my QA on your other recent work, so I went ahead and tested this as well. Worked great on both macOS and Windows!

mphstudios (approved):

PR #1060, opened 2025-09-12
Fix windows path separator leaking into epub asset paths (DEV-21534)

cbutcosk:
This PR updates the epub transform to resolve an issue on where Windows-style paths (`example\test.png`) were leaking into the epub representation in the generated epub package. Changes:
- Refines logic for moving epub assets to only warn if an asset is not present.
- Updates test docs, adds epub file to CI artifacts,  and adds `test:clean` script to remove all test artifacts
- Tweaks eleventy environment mocking function to take more involved globalData properties
- Add test for epub resources that leak Windows path separators

PR #1061, opened 2025-09-12
Remove slugging on external links in PDFs (DEV-20818)

cbutcosk:
This PR resolves an issue where anchor links in the PDF representation were being slugged. It adds a test and resolves it by adding a check in the relative link transform used by the pdf transform. It is based on the updates to the eleventy environment mocking function in #1060 , which should be reviewed and merged first.


mphstudios (dismissed):

geealbers (dismissed):
Confirming here for the record that the fix works for me testing locally on Windows and macOS

mphstudios (approved):

mphstudios (approved):

Issue #853, opened 2023-10-06
`{% contributors context=publicationContributors %}` is causing contributors created with `first_name` `last_name` in page YAML to be overwritten


Erin-Cecele:
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [X] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

When using the `{% contributors %}` shortcode with `context=publicationContributors` all the contributors mentioned in page YAML and publication.yaml should be listed on the page with the shortcode. Under each contributor it should list the essays written by each contributor. 

### Actual Behavior

When using the `{% contributors %}` shortcode with `context=publicationContributors` the shortcode is looking for contributors listed in `publication.yaml` as well as any listed in the page YAML of individual markdown pages. But for those listed in page YAML, the data for each one is overriding the last, except for the links to papers. So, you end up with the info of the last contributor to be listed in the book, but with a list of every paper that all the other overridden contributors were part of.

### Steps to Reproduce

1. Create a new project
2. Add two different contributors to the page YAML of two different `.md` files using the `first_name` and `last_name` fields. 
3. Look at the Contributors page of project and you will see only **one** of the two contributors listed and both essays listed under that one contributor 

### Version Numbers

[Project]
 quire-cli 1.0.0-rc.10
 quire-11ty 1.0.0-rc.14
 starter /Users/edunigan/GitHub/quire-starter-minimal@2.6.0
[System]
 quire-cli 1.0.0-rc.10
 node v18.16.0
 npm 9.5.1
 os Darwin 21.6.0

### Web Browser

Chrome Version 117.0.5938.88 (Official Build) (x86_64)

### Relevant Terminal/Shell Output

```shell
N/A
```

### Supporting Information

There are two separate Forewords in this project with two separate contributors, but only one contributor is listed with both Forewords.

<img width="643" alt="Screen Shot 2023-10-06 at 1 55 41 PM" src="https://github.com/thegetty/quire/assets/62075223/886c3baa-199c-4faa-8414-fc57fdf657c7">

Erin-Cecele:
This is similar to, but a slightly different application from issue #847

Erin-Cecele:
When I refer to "contributors mentioned/listed in the page YAML" I am specifically referring to the contributors added with: 

```
contributors
  - first_name: 
    last_name: 
```

This is separate from contributors listed in the `publication.yaml` and added via an `id` in the page YAML, which this issue does not appear to affect.

[...]

Issue #847, opened 2023-09-14
`{% contributors context=pageContributors format='bio' %}` lists all publication essays, not just for `pageContributors`


Erin-Cecele:
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [X] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

{% contributors context=pageContributors format='bio' %} should list all the contributors included in Page YAML, their individual bios, and their other essay contributions within the publication _unless that contribution is the page the shortcode is being used on_.

### Actual Behavior

{% contributors context=pageContributors format='bio' %} lists the contributors included in that page's page YAML, their bios, and, erroneously, all the other essays by page YAML contributors throughout the publication as well as the essay the shortcode appears on.

<img width="653" alt="Screen Shot 2023-09-14 at 1 20 30 PM" src="https://github.com/thegetty/quire/assets/62075223/1b0b26d5-2b52-4c6b-a2e8-ed0ae67f1a79">

In the example above, Mahdi Sahragard only contributed the first essay. None of the other items in that list should be appearing.

### Steps to Reproduce

1. Add `contributor` info to the Page YAML of an `essay`
```
contributor: 
  - first_name: Anneka  
    last_name: Lenssen
    bio: 
```
2. Add the shortcode `{% contributors context=pageContributors format='bio' %}`

### Version Numbers

[grj-19]
 quire-cli 1.0.0-rc.10
 quire-11ty 1.0.0-rc.14
 starter /Users/galbers/GitHub/quire-starter-default@2.6.0
[System]
 quire-cli 1.0.0-rc.10
 node v18.16.0
 npm 9.5.1
 os Darwin 21.6.0

### Web Browser

Chrome: Version 117.0.5938.62 (Official Build) (x86_64)

### Supporting Information

I found that adding *** above the shortcode removed the list of essays:

```
***
{% contributors context=pageContributors format='bio' %} 
```

Erin-Cecele:
This is the same as issue #853 - an issue inspired by an error in Inventories and Surveys: https://github.com/thegetty/inventories-and-surveys/issues/9

Erin-Cecele:
When I refer to "contributors included in that page's page YAML" I am referring to the contributors added with: 

```
contributors
  - first_name: 
    last_name: 
```

This is separate from contributors listed in the `publication.yaml` and added via an `id` in the page YAML.

PR #1094, opened 2025-11-18
Fix page data contributor regression in 1.0.0-rc.39

cbutcosk:
This PR fixes regressions in quire-11ty@1.0.0-rc.39 where publications with a page with no `contributor` propery in page headmatter would crash builds, multiple contributors without ids would trigger duplicate element id errors (on `id='undefined'`), and contributors from page data with the same first_name + last_name pairs were not being properly deduplicated. Details:
- Adjusts `contributors` shortcode test cases to better cover state + error space
- Uses slugged first_name-last_name string as fallback id for contributors without one
- Corrects contributor coalescing across pages in `publicationContributors`
 
These issues are tracked by https://github.com/thegetty/quire/issues/853#issuecomment-3534338465 and https://github.com/thegetty/quire/issues/853#issuecomment-3577118911

### DIFF
--- a/packages/11ty/_plugins/shortcodes/contributors.js
+++ b/packages/11ty/_plugins/shortcodes/contributors.js
@@ -95,6 +95,8 @@ export default function (eleventyConfig) {
       case 'name-title-block': {
         const separator = (format === 'name-title') ? ', ' : ''
         const listItems = contributorList.map((contributor) => {
+          const { first_name: firstName, id, last_name: lastName } = contributor
+
           const contributorParts = [
             `<span class="quire-contributor__name">${fullname(contributor)}</span>`
           ]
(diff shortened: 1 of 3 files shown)


mphstudios (approved):

geealbers:
@cbutcosk sorry, I tested this earlier today and found an issue with it throwing duplicate ID errors, but then commented about it in the wrong PR. Please see https://github.com/thegetty/quire/issues/853#issuecomment-3577118911

geealbers:
@cbutcosk I can confirm that the duplicate id error caused by using two or more contributors defined with `first_name` and `last_name` in the page headmatter is now fixed. However, I'm now seeing that it also throws duplicate id errors when I try to use `{% contributors context=pageContributors format='bio' %}` on a page. Sorry!

Using the shortcode on the page like that would be a common use case for something like a journal article, where you'd want the contributor name(s) in the page header, and then would want to include their full bios at the bottom of the page.

Fixing this could also be an extension of issue #847 if you want to merge this PR in now and then do a separate one for this newly discovered issue.

Issue #1068, opened 2025-09-25
Zip files generated in home directory when new project is created


Erin-Cecele:
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [x] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

Only the new project folder should be installed in the home directory when a new project is generated with `quire new project-name`.

### Actual Behavior

Zip files are being generated in addition to the new project folder. 

This image represents new projects generated with three different versions of Quire.

<img width="723" height="66" alt="Image" src="https://github.com/user-attachments/assets/f735218d-1335-4928-abb6-194aeee1ae72" />

If you create more than one new project on a specific version of Quire the older zip file will be written over. 

### Steps to Reproduce

1. use Node v. 22
2. run the command `quire new test-project --quire-version 1.0.0-rc.36-issue1066`
3. navigate to your home directory and notice the new zip folder

### Version Numbers

edunigan@GT35063 test-project % quire info --debug
[CLI] Command 'info' called with options { debug: true }
[test-project]
 quire-cli 1.0.0-rc.33
 quire-11ty 1.0.0-rc.36-issue1066
 starter https://github.com/thegetty/quire-starter-default@2.9.0
[System]
 quire-cli 1.0.0-rc.33
 node v22.18.0
 npm 10.9.3
 os Darwin 23.6.0
edunigan@GT35063 test-project % 

### Web Browser

N/A

### Relevant Terminal/Shell Output

```shell
N/A
```

geealbers:
Done with pr #1211

PR #1106, opened 2025-12-18
Remove tarball after use in `quire new` command

cbutcosk:
This PR adds an integration test for https://github.com/thegetty/quire/issues/1068 , adjusts the `quire new` command to remove `@thegetty/quire-11ty` tarballs downloaded during publication install, and updates the underlying filesystem calls of `installInProject()` to use native node API calls for better testability.

### DIFF
--- a/packages/cli/src/commands/create.test.js
+++ b/packages/cli/src/commands/create.test.js
@@ -1,7 +1,8 @@
+import esmock from 'esmock'
+import path from 'node:path'
+import sinon from 'sinon'
 import test from 'ava'
 import { Volume, createFsFromVolume } from 'memfs'
-import sinon from 'sinon'
-import esmock from 'esmock'
 
 test.beforeEach((t) => {
   // Create sinon sandbox for mocking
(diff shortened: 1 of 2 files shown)


mphstudios (approved):

packages/cli/src/commands/create.test.js
```diff
@@ -292,3 +293,44 @@ test('create command should pass quire-path option to methods', async (t) => {
...
+    },
+    'execa': {
+      // Mock `npm pack` behvaior of creating a file in .temp
+      execaCommand: sandbox.stub().withArgs(/npm pack/).callsFake(() => {
+        fs.writeFileSync(path.join(projectRoot, '.temp', 'thegetty-quire-11ty-1.0.0.tgz'), '')
+      })
+    }
+  })
+
+  await quire.installInProject(projectRoot, '1.0.0')
```
mphstudios: 💭 the `lib/quire` module might provide its own **simple** mock that abstracts the internal concerns of calling `installInProject`

💭 implementing a `lib/npm` module that wraps the `npm` commands (`init`, `install`, `uninstall`, `pack`, `clean`) would provide a consistent way for the other modules to interact with `npm` and abstract internal concern for how it is called (`execa` vs node `child_process`)

mphstudios (approved):

PR #1133, opened 2026-01-14
Refactor cli path handling

mphstudios:
Refactor path handling for interactions with Eleventy. 

## Purpose

The current eleventy path configuration is a "leaky abstraction" that uses string literal templates to compose paths whenever needed. This change consolidates to path configuration in an `11ty/Paths` class, which exports accessor methods for paths.

**_See the [README](https://github.com/thegetty/quire/pull/1133/changes#diff-0753ceb9a69be2b7051736f4bfc1c4c39acc06c86a060b2783ed0f23e027482c) included in this pull request for detailed documentation of these changes._**

**Nota bene**
The current pattern for handling `paths` is still a rather "leaky abstraction" because Quire _is an Eleventy based tool_; the Eleventy `paths` and general _project_ `paths` are mixed. This "leak" might be resolved in future refactoring that aligns the CLI with domain concerns: `paths` are a concern of a `Project`.

### DIFF
--- a/packages/cli/src/lib/11ty/api.js
+++ b/packages/cli/src/lib/11ty/api.js
@@ -3,6 +3,33 @@ import path from 'node:path'
 import paths from './paths.js'
 
 /**
+ * Configure environment variables required for Eleventy.
+ *
+ * These must be set before the Eleventy instance is created and the
+ * `.eleventy.js` configuration file is parsed.
+ *
+ * @see https://github.com/11ty/eleventy/issues/2655
+ *
+ * @param {Object} options
+ * @param {'production'|'development'} options.mode - Build mode
+ * @param {boolean} options.debug - Enable Eleventy debug output
+ */
+const configureEleventyEnv = ({ mode = 'production', debug = false } = {}) => {
+  // Path configuration for decoupling quire-11ty from project input directory
+  process.env.ELEVENTY_DATA = paths.getDataDir()
+  process.env.ELEVENTY_INCLUDES = paths.getIncludesDir()
+  process.env.ELEVENTY_LAYOUTS = paths.getLayoutsDir()
+
+  // Build mode
+  process.env.ELEVENTY_ENV = mode
+
+  // Debug output
+  if (debug) {
+    process.env.DEBUG = 'Eleventy*'
+  }
+}
+
+/**
  * A factory function to configure an instance of Eleventy
  * @see https://www.11ty.dev/docs/config/#configuration-options
  *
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/11ty/paths.js
```diff
@@ -6,98 +6,179 @@ const __filename = fileURLToPath(import.meta.url)
...
+
+  // ─────────────────────────────────────────────────────────────────────────
+  // Utility methods
+  // ─────────────────────────────────────────────────────────────────────────
+
+  /**
+   * Get all paths as an object (for backward compatibility and logging)
+   * @returns {Object} Object containing all path values
+   */
+  toObject() {
```
cbutcosk: In practice this method name obscures the semantics of what a caller gets out of the singleton  -- something like `allPaths()`, `getAll()`, `pathsObject()`, etc, that would serve the same purpose with more clarity for later readers?

packages/cli/src/lib/11ty/README.md
```diff
@@ -16,20 +16,86 @@ See [Eleventy Documentation: Command Line Usage](https://www.11ty.dev/docs/usage
...
+
+Eleventy 3.x provides `setIncludesDirectory()` and `setLayoutsDirectory()` API methods, but environment variables are preferred because:
+
+1. **cli.js compatibility** - The `cli` module runs Eleventy as a subprocess via `execa`. API methods cannot be called on a subprocess; only arguments and environment variables can be passed.
+
+2. **Same path restriction** - The API methods have the same limitation: paths must be relative to the input directory. Using `../_includes` works identically with either approach.
+
+3. **Consistent implementation** - Environment variables work for both `api.js` and `cli.js`, avoiding code divergence between the two modules.
+
+**Known limitation:** The Eleventy `TemplatePathResolver` requires layouts and includes directories to be within the input directory, which prevents fully decoupling Quire project content from `quire-11ty` code. See [eleventy#2655](https://github.com/11ty/eleventy/issues/2655) (still open).
```
cbutcosk: I'm confused by this limitation note -- like the other users on that issue thread I have not had any trouble fully separating input content and layouts + includes?

cbutcosk (changes requested):
I noted two small things worth looking at -- the method `toObject` could use a slight rename and I would remove the closing note about pathing limitations of 11ty, which haven't been true in my experience

packages/cli/src/lib/11ty/paths.js
mphstudios: Agreed, though I have started some refactoring stacked on these changes that will address aligning the paths with domain concerns.

cbutcosk: Renaming `toObject` in another PR is fine

cbutcosk (approved):

PR #1211, opened 2026-03-30
Add test for removal of tarball and temp dir, remove tarball

cbutcosk:
This PR restages #1106 

Original description:
---- 
This PR adds an integration test for https://github.com/thegetty/quire/issues/1068 , adjusts the quire new command to remove @thegetty/quire-11ty tarballs downloaded during publication install, and updates the underlying filesystem calls of installInProject() to use native node API calls for better testability.

PR #1210, opened 2026-03-30
Add pagefind search

cbutcosk:
This PR incorporates the PageFind search feature from https://github.com/thegetty/quire/pull/1084 .

PR #1212, opened 2026-03-30
Silence various sass deprecation warnings

cbutcosk:
This PR restages #1129 against `main`'s rewound commit history.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,6 +119,7 @@ commands:
       - run:
           name: Run integration tests
           command: mkdir -p reports && npm run test:integration
+          no_output_timeout: 15m
 
       # Persist the site build for browser testing and submit test reports
       - persist_to_workspace:
--- a/packages/11ty/_includes/components/lightbox/styles.js
+++ b/packages/11ty/_includes/components/lightbox/styles.js
@@ -23,12 +23,12 @@ export default function (eleventyConfig) {
     const sassOptions = {
       api: 'modern-compiler',
       loadPaths: [path.resolve('node_modules')],
+      quietDeps: true,
       silenceDeprecations: [
         'color-functions',
         'global-builtin',
-        'import',
-        'legacy-js-api',
-        'mixed-decls'
+        'if-function',
+        'import'
       ]
     }
     lightboxCSS = sass.compile(lightboxStylesPath, sassOptions)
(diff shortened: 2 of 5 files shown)


geealbers:
@cbutcosk Oddly, this seems to have brought back SASS deprecation warnings that I wasn't seeing when testing PR #1210 and #1211?

geealbers:
@cbutcosk This is weird. It works fine on macOS but on Windows I'm running into what I think should be an unrelated error when running `quire preview` and one I didn't run into when testing the last two PRs:

```
[11ty] Copied 114 Wrote 0 files in 12.03 seconds (v3.1.0)
[11ty] Eleventy Fatal Error (CLI):
[11ty] 1. Having trouble writing to "./_site/about/index.html" from "./content/about.md" (via EleventyTemplateError)
[11ty] 2. Transform `@11ty/eleventy/html-transformer` encountered an error when transforming ./content/about.md. (via EleventyTransformError)
[11ty] 3. You have more than one HTML `id` attribute using the same value (id="section-learn-more") in your template (./content/about.md). You can disable this error in the IdAttribute plugin with the `checkDuplicates: false` option.
[11ty]
[11ty] Original error stack trace: Error: You have more than one HTML `id` attribute using the same value (id="section-learn-more") in your template (./content/about.md). You can disable this error in the IdAttribute plugin with the `checkDuplicates: false` option.
[11ty]     at file:///C:/Users/geeal/test-pr-1212/node_modules/@11ty/eleventy/src/Plugins/IdAttributePlugin.js:69:15
[11ty]     at C:\Users\geeal\test-pr-1212\node_modules\posthtml\lib\api.js:85:52
```

I started the test project with:

```
quire new test-pr-1212 --quire-path C:\Users\geeal\Downloads\quire\packages\11ty
```

The `id="section-learn-more"` the error message is referencing is coming from an accordion shortcode on the about.md page. Setting `checkDuplicates: false` in `.eleventy.js` does clear the error and at that point I can see that the SASS deprecation warnings have been properly silenced. Thoughts?

cbutcosk:
@geealbers Hmm I don't see that problem on virtualized windows when I test it and it's passing on actual windows in CI. Furthermore if I recall those attributes should now be prefixed with `epub-` / `pdf-` or possibly written out (can't recall off hand).

We've used this branch name before, I wonder if your clone is outdated... When you have `fix/sass-deprecations` checked out could you try `git reset --hard origin/fix-sass-deprecations` and making a new publication as you had been? And / or try a fresh clone of the repo (checked out to the right branch) for passing in to `quire new --quire-path ...`?

geealbers (approved):
omg, so dumb of me. I `npm link`ed to the wrong copy of the `quire` repo on my tester PC. Thank you for the clue that got me to realize that. Now that I'm on the correct version, I can confirm this PR works as intended and can be merged. Thanks!

PR #1193, opened 2026-02-09
Feature: Add q-search-results-list web component.

fchasen:
### Description

Implements a new `q-search-results-list` web component to display and manage the searching a Pagefind index, replacing the current search results template and display logic.

- The component takes a `query` attribute to search for in a Pagefind index.
- Removes the `updateSearchResults` and `displaySearchResults` methods, as lit will handle batching UI updates.
- Enables adding figures to the Pagefind index.
- Adds subresults, contributors metadata and figures with image previews to the search results.
- Adds excerpts with search query highlighting to all search result items.
- Removes type and word count metadata from the search results.

### Checklist 

- [X] I have read the [CONTRIBUTING.md](https://github.com/thegetty/quire/blob/main/CONTRIBUTING.md) file

- [X] I have made my changes in a new branch and not directly in the main branch

- [x] This pull request is ready for final review by the Quire team

### Include screenshots of before/after if applicable.

quire-search

### Additional Comments

Thumbnail handling for figures will be added once they are a size that can be used.

PR #1220, opened 2026-04-23
Process all images in figure plugin, log figures according to debug flag

cbutcosk:
This changeset updates quire's figure handling to produce thumbnail, static inline, and print images for all figure images. It also connects the CLI's `--debug` flag to the figure plugin for log level adjustments and clarifies plugin-internal architectural intent by renaming the internal `Figure` class and its associated factory and `get` filter to `FigureMedia`, `FigureMediaFactory`, and `getFigureMedia`.   

This PR finishes quire-side work necessary to merge the searchable figures feature in #1193 .

## Feature changes

- Adjusts code paths to process dimensions and derivative images for all figure images, adds tests.
- Adds `height` and `width` attributes to the `img` tag in the figure component when it is used for figures in the running text.
- Updates 11ty configuration to accept debug log from a `QUIRE_DEBUG_LOG` environment variable. Updates CLI to set `QUIRE_DEBUG_LOG` when called with `--debug` flag.

Use `quire build --debug` or `quire preview --debug` to restore figure logging, or set the `QUIRE_DEBUG_LOG` env var to accomplish the same thing without triggering 11ty debug logging, ala `QUIRE_DEBUG_LOG=1 quire build` /  `QUIRE_DEBUG_LOG=1 npm run build`.

## Implementation notes

- Plugin code intent suffers from using "figure" to mean data from user-supplied `figures.yaml` and data produced from processing figure media files. This PR refactors the plugin to use "figure to mean user-supplied data and "figureMedia" to mean data objects created by the plugin. Accordingly the class `Figure` is now `FigureMedia`, created by a `FigureMediaFactory`, data consumers use a `getFigureMedia` filter to fetch data, and plugin data is stored in the `figureMedia` 11ty global data property.


geealbers:
@cbutcosk All my QA testing (on both macOS and Windows) is done and clear. Just a few things I wanted to query:

I like the addition of image processing for non-IIIF figures. I noticed, however, that along with the processed files (print, static-inline, thumbnail) there is also a duplicate of the original image file included in the `/iiif/` directory. For example, there is now both `_site/iiif/fig-1/evans-legionnaire.jpg` and `_site/_assets/images/figures/evans-legionnaire.jpg` included in the build output, but only the one in `_assets/images/figures/` seems to be used anywhere. Can we eliminate of these new, extra copies?

Then, I just had two naming questions. I want us to be particularly intentional on how things are named and organized and/or I personally want just to understand our naming choices better. I'm thinking ahead to future developer documentation and our ongoing focus on utility, **clarity**, and reliability for Quire.

1. You've used `FigureMedia` in place of `Figure` here. I take that to indicate the work is about the actual media (the `img`) of the figure, rather than the figure as a whole, which would also include the figure caption and its elements. Is that a correct assumption? Is there more to the name change than that?
2. We're now processing non-IIIF image assets and placing them into the `_site/iiif/` directory. And the image processing itself is configured in the `_plugins/figures/iiif` directory. I understand the expedience of it, but should we think of renaming and/or reorganizing these into different directories? Eventually, if not now? I'm just imagining that someone coming into Quire cold might otherwise find the conflation confusing.

cbutcosk:
@geealbers Yeah the naming change is mostly to make it easier to work in the FigureMedia class. I anticipate some more cleanups to come so it can be a little less difficult to work in the figures plugin. For example #1235 addresses your second question by precomputing paths to the new derivatives and using them in components. To your specific questions:

1. Your reading is correct, this is just naming changes so it's clearer what is happening. A common design problem in quire is that it elides the difference between user-supplied types (Figure, Object, Page, etc) and the publication domain types quire uses to do its technical business (FigureMedia, ObjectMetadata, PageContent, etc). The user-supplied data is still passed on to consuming components in addition to the result data from processes that quire does to the figure media.

    In quire this change is usually not modeled as a class but the figures plugin does so, and manages enough complexity that the elision between user types and domain types makes following the code flow difficult. For example (to pick on one I wrote) the figure image tag does some desperate force-removing of `\` characters that have leaked into paths because the design assumes that user-supplied figure data is more or less transparent to the publication-supplied asset paths. Overall the goal of this and the other figure-centric PR is to allow for deleting some of the needless complexity in figure components.

2. The first part migrating away from the `/_assets/images/figures/...` form of image asset paths (I agree vehemently!) is updating figure components to the use these newly generated assets. #1235 does that in draft, it should be ready sometime next week. I can trial omitting processed figure images from the passthrough copy of the figures directory there, I suspect it should be fine through other asset types (video, possibly tables) still require the bulk copy.

geealbers:
@cbutcosk thanks so much for those explanations!

geealbers (approved):

Issue #1197, opened 2026-02-12
Transform global data (figure.src)


geealbers:
### Before proceeding, check to make sure there isn’t an existing issue for this idea or a similar one.

- [x] I have searched the existing issues and determined this is a new proposal.

### Proposed change

This is an enhancement for our sanity and not repeating logic. There should be a way to prepend all relative values for figure.src with the image directory when the project builds, so that we don't have to repeat this logic in multiple places every time a figure src is referenced in a shortcode (currently in: figure image shortcode, figure video shortcode, get-image-service helper)

Possible solutions:

1.) Create a computed figures property in eleventyComputed

2.) Add a method to add optional transforms to data before figures is added to the global data in _plugins/globalData

### Potential Impact

Sanity

### Supporting Information

This suggestion was originally posted to Jira on 15/Aug/22 by Hannah as [DEV-12941](https://jira.getty.edu/browse/DEV-12941).

As discussed, @cbutcosk will be looking at it as part of his work on issue #1171.

cbutcosk:
The remaining aspects of this task are:
- Specifying tests for figures plugin ensuring that user-supplied image types in data like contributor avatars, promo images, and publisher logos are processed into figureMedia objects. No _transformations_ are necessary on these images but their dimensions and paths should be available and their full-resolution images should be copied into built publication bundles. See #1279 .
  - `epub.defaultCoverImage`
  - `publisher[].logo`
  - `contributor[].image`
  - `promo_image`
- Specifying tests for figures plugin ensuring figure properties like poster image paths and video or audio asset paths are processed. See #1285 .
  - `poster` - should be transformed for use in print, etc contexts
  - `src` when mediaType == `video` the plugin should calculate the asset paths
  - `src` when mediaType == `audio` the plugin should calculate the asset paths
  - `annotations[].items[].src` - should be transformed as necessary
  - `annotations[].items[].src` - should be transformed as necessary
- Implementation of figures processing for the above ancillary and primary asset paths -- video, audio, and image files necessary should all be moved in-place. These should be handled automatically by vite during the asset bundling.
- Refactoring away component-calculated asset path for bio, page header, etc components. See #1279 and #1285 .

Issue #1171, opened 2026-01-26
All figures should have a correctly defined thumbnail with sensible fallbacks


geealbers:
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [x] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

Every figure should have a `thumbnail` defined and every `thumbnail` should correctly resolve to an existing image.

If `thumbnail` is defined for the figure in `figures.yaml` it should be used; if not, but if the figure is set to `zoom: true` the thumbnail generated by sharp should be used (see `_plugins/figures/iiif/config.js`); otherwise, the `src` in `figures.yaml` should be used.

A slightly more fulsome solution here would also include fallbacks for figures that don't have viable `src` paths like tables, sequences, and annotations images. But at least with the schema above, users have a way of manually defining thumbnails for those specialty figures.

### Actual Behavior

Due to unfinished work, the `thumbnail` for all figures is being set to the `staticInlineFigureImage`, which only exists for IIIF images. See lines 350–51 in `_plugins/figures/figure/index.js`.

```js
// TODO: implement thumbnail getter
thumbnail: this.staticInlineFigureImage
```

### Steps to Reproduce

n/a

### Version Numbers

All 1.0.0 versions up to and including `quire-cli 1.0.0-rc.33` and `quire-11ty 1.0.0-rc.40`

### Relevant Terminal/Shell Output

```shell

```

cbutcosk:
@geealbers With the recent merges and work upcoming on #1197 and #1191 the only missing part of this issue will be overriding `thumbnail` with user-supplied data in YAML. _However_ the API for asset paths and metadata of derivatives looks a little different than this issue is expecting, in that there are thumbnail, static inline, and print images being generated. Should users be able to override any / all of those with a same-named property in the YAML?

geealbers:
@cbutcosk as discussed, I don't think we need to allow users to override the thumbnail specifically. Instead I think we should focus on the `poster` option as suggested in #1191. If users can define a `poster` image in the YAML and Quire can create and use derivatives from that (thumbnail, static inline, and print), that seems a clean solution useful for annotation images where the issue is most acute, and it would be consistent with how video images work as well. I could also see the same being useful for sequence images, audio, and maybe even `media: table` (which I sometimes use as a hack to include other kinds of HTML embeds as figures).

Make sense?

cbutcosk:
[...]

PR #1235, opened 2026-05-12
Store figure image derivative dimensions and paths on `derivatives` property

cbutcosk:
This PR refactors the figure media model to make it easier to use image derivatives generated during publication builds. It moves path and dimension metadata handling into the higher-order `FigureMedia` class, allowing for more correct re-use of the paths in components and adding long-missing `height` and `width` attributes to quire's figure `img` tags, improving page responsiveness.

## Overview of changes:
- Adds a `derivatives` property to the `media()` data shim on `FigureMedia` class. `derivatives` contains an object with derivative transformation names as keys (full, thumbnail, etc) and result metadata (height, width, and asset paths) objects. A full list of transforms is in `packages/11ty/_plugins/figures/iiif/config.js`.
- This is the shape of the additional data on `FigureMedia` instances serialized as JSON:
```json
{
  "derivatives": {
    "full": {
      "dimensions": {
        "height": 1000,
        "width": 1000
      }
      "paths": {
        "absolute": "/_assets/images/XXX.jpg",
        "internal": "/subpath/_assets/images/XXX.jpg",
        "uri": "https://example.org/subpath/_assets/images/XXX.jpg"
      }
    }
  }
}
```
- `paths` contains asset paths for derivatives:
  - `absolute` - relative to the the root of the publication hostname
  - `internal` - relative to the root of the publication (ie, paths that are safe to use in fragments that will be processed by the asset bundling plugin)
  - `uri` - with scheme, hostname, and port
- `dimensions` contains the `height` and `width` for the derived image
- Moves path calculations out of components and into `FigureMedia` methods. The switching dispatch of the `figureElement` component is also refactored so that control flow branches within the static figure tag are reduced from several branches to one branch with a varying argument.
- Deletes now-redundant path calculations from `imageTag` and the print output mode of `figureImage` components.
- Adds a helper function, `snakeCaseToCamel`, for converting transformation names within in the IIIF configuration that are in snake case to the camel case used by quire's data APIs (and enforced by linting, etc). 
- Adds tests for image asset paths to ensure correct behavior with and without a publication pathname.
- Adds test ensuring that figure images with an HTTP URL as `src` but not an internally-served IIIF URL (eg, images on a CDN) use the src URL for all paths.

## Affected Issues:
- These changes address the images side of #1171, transformed derivatives are generated for all images and in this PR their paths and metadata are made available at `globalData.figuresMedia[].derivatives.thumbnail.paths`, etc.
- Also addressed is the images part of #1197: this PR hoists pathing logic out of the html figure image component and into the class.
- Still to be addressed in future PRs are video and audio asset paths, poster paths, and other asset-encumbered figure media types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1298,7 +1298,6 @@
       "version": "4.6.1",
       "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.6.1.tgz",
       "integrity": "sha512-KTsJMmobmbrFLe3LDh0PC2FXpcSYJt/MLjlkh/9LEnmKYLSYmT/0EW9JWANjeoemiuZrmogti0tW5Ch+qNUYDw==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "eslint-visitor-keys": "^3.4.3"
@@ -1317,7 +1316,6 @@
       "version": "4.12.1",
       "resolved": "https://registry.npmjs.org/@eslint-community/regexpp/-/regexpp-4.12.1.tgz",
       "integrity": "sha512-CCZCDJuduB9OUkFkY2IgppNZMi2lBQgD2qzwXkEia16cge2pijY/aXi96CJMquDMn3nJdlPV1A5KrJEXwfLNzQ==",
-      "dev": true,
       "license": "MIT",
       "engines": {
         "node": "^12.0.0 || ^14.0.0 || >=16.0.0"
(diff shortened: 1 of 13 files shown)


geealbers:
@cbutcosk I ran this through my normal QA process and everything is looking good. 

1. Is there any documentation that should be added to `_plugins/figures/README.md`, `_plugins/figures/schema.json` or any other location in the code base to capture the information in your description of this PR or just to document things for our future selves?

2. I'm going to post new issues about [removing the duplicate figure image files from build that I surfaced in the previous PR](https://github.com/thegetty/quire/pull/1220#issuecomment-4454049636), and about adding values to config.yaml for the derivative image sizes. Both for follow-up at a later date. Anything else you'd like to add as a follow-up issue that we might want to address in the future?

geealbers (approved):
Sorry for the delay in reviewing @cbutcosk, but this all passes QA. Thanks!

Issue #1224, opened 2026-05-01
Cover and splash page background images are not displaying in PDF


geealbers:
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [x] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

Cover and splash page images will display in the PDF output

### Actual Behavior

On macOS, background images for the cover and splash pages are appearing in pdf.html with `&quot;` escapes:

```html
background-image: url(&quot;/_assets/images/spiral-overlay.png&quot;);

background-image: url(&quot;/_assets/images/figures/lange-house.jpg&quot;);
```

I believe as a result, the `trimLeadingSeparator` function in `_plugins/transforms/outputs/pdf/transform.js` isn't working and the images are not displaying in the PDF output.

### Steps to Reproduce

1. Start a new project with `quire-11ty@1.0.0-rc.44`
2. Run `quire build` and `quire pdf`

### Version Numbers

[test-1.0.0-rc]
 quire-cli 1.0.0-rc.52
 quire-11ty 1.0.0-rc.44
 starter https://github.com/thegetty/quire-starter-default@2.10.0
[System]
 quire-cli 1.0.0-rc.52
 node v22.22.0
 npm 10.9.4
 os Darwin 22.6.0

### Relevant Terminal/Shell Output

```shell

```

geealbers:
fwiw, when this issue appeared, I spent a little time looking at getting rid of these css background-images on the cover and splash pages entirely as they often cause problems in our outputs and they're not accessible. In the short-term, it's relatively easy to do on the Cover layout, but messier on the Splash page. On splash, it opened up questions for me about why we don't use the standard `page-header.js` template there and why we have two different layouts at all. Long-term I'd like to see us simplify the code and move to using `presentation: splash` that would have a sensible/consistent effect on any of the layouts, but that's a project for another day.

geealbers:
This is also an issue in the EPUB output, where background images are appearing as:

```
style="&#xA;                background-image: url(&quot;/_assets/images/spiral-overlay.png&quot;);&#xA;              "
style="&#xA;              background-image: url(&quot;/_assets/images/figures/lange-house.jpg&quot;);&#xA;            "
```

This is some EPUB validation issues that we'll need to make sure are cleared once the images are appearing correctly:

```
Validating using EPUB version 3.3 rules.
[...]

PR #1250, opened 2026-06-03
Replace css background images with html img elements

geealbers:

*For the Title of this pull request, please use the format "Type/Issue-#: Brief description." For Type, the options are Fix, Feature, Docs, or Chore. Issue-# is only needed if this pull request addresses an [existing issue](https://github.com/thegetty/quire/issues).*

**Before submitting your PR, make sure that you can run the following commands: `quire preview`, `quire build`, `quire pdf`, and `quire epub`. If any of those core quire commands throw errors/yield unexpected behavior, we will NOT review the PR!**

### Checklist 

- [X] I have read the [CONTRIBUTING.md](https://github.com/thegetty/quire/blob/main/CONTRIBUTING.md) file

- [X] This pull request is ready for final review by the Quire team

Issue #1224 ("Cover and splash page background images are not displaying in PDF")

This part of a fix for the issue. It is dependent on [PR #54](https://github.com/thegetty/quire-starter-default/pull/57) in the in the `quire-starter-default` repo to update the SCSS.

Replaces the use of CSS background images with HTML img elements.

No

### DIFF
--- a/packages/11ty/_layouts/cover.liquid
+++ b/packages/11ty/_layouts/cover.liquid
@@ -5,17 +5,15 @@ description: Quire publication cover page
 layout: base.11ty.js
 ---
 
+{% if image %}
 {% assign coverImage = image %}
 {% assign imagePath = config.figures.imageDir | concat: coverImage | join: '/' %}
+{% endif %}
 
 <section class="quire-cover__hero hero is-fullheight">
-  <div
-    class="quire-cover__overlay"
-    style="background-image: url('{{ imagePath }}');"
-  >
-    {% comment %} This ensures background image asset gets copied into epub package {% endcomment %}
-    <img class="visually-hidden" src="{{ imagePath }}" alt="" data-outputs-include="epub" />
-  </div>
+  {% if image %}
+    <img class="quire-cover__hero-image hero-image" src="{{ imagePath }}" alt="" />
+  {% endif %}
   <div class="quire-cover__hero-body hero-body">
     <div class="container is-fluid">
       <h1 class="title" id="page-header-{{ page.filePathStem }}">
--- a/packages/11ty/_layouts/splash.liquid
+++ b/packages/11ty/_layouts/splash.liquid
@@ -4,15 +4,14 @@ classes:
 layout: base
 description: splash page layout
 ---
-{%- if image -%}
+{% if image %}
   {% assign imagePath = config.figures.imageDir | append: '/' | append: image %}
 {% endif %}
 {% assign labelDivider = config.pageTitle.labelDivider %}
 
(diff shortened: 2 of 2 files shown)


cbutcosk:
@geealbers I'm reviewing here even though some of these issues are probably changes needed in the starter repo..
- When I use this update to make a PDF with pagedjs, the colors of the cover image look a little dull and the image is fairly jaggy. Is there maybe a missing height constraint or a missing `object-fit`?
Screenshot 2026-07-09 at 10 40 59

- When viewing the resulting epub in ibooks at the width where it shows two-pages, the intro page gets cut off by the skeumorphic book spine. Assuming this has been happening before this change, it did make me wonder if there are epub options re: layout that might help when added at some later date.  
Screenshot 2026-07-09 at 10 41 37

The PDF issue seems like the only blocker to merge (unless that's a known issue? didn't remember it from looking at spec PDFs)

geealbers:
Thanks @cbutcosk.

I was able to fix the EPUB issue by switching the image wrapper from `width: 100vw;` to `width: 100%;`. I think I'd only been testing in a 1-page mode in the my EPUB reader and hadn't stopped to think that being in a 2-page mode would change the way viewport width related to page width.

Screenshot 2026-07-09 at 10 38 45 AM

I think there are a couple things going on with the cover image in the PDF. First, having switched from a true CSS gradient with transparent PNG to a static JPG, we lost some of the crispness of the lines and smoothness of the gradient. We're also now only seeing the middle part of the gradient whereas before we would have seen the brighter starting colors on either side. When/if we rethink/redesign the starter content in the future we'd be better served using a less finicky cover image.

That said, I also found that the Paged.js output was noticeably less saturated / grayer than the PrinceXML output! In the screenshot below, the Paged.js version is on the left, the PrinceXML version (which matches the source image) is on the right. 

Screenshot 2026-07-09 at 9 51 39 AM

It turns out, this was because the image was saved with an Adobe RGB (1998) color profile, which apparently Paged.js doesn't like? When I saved it with a sRGB IEC61966-2.1 color profile instead, the outputs of both matched each other and matched the source image.

I've pushed these fixes to [PR #57](https://github.com/thegetty/quire-starter-default/pull/57) on `quire-starter-default`. So, that PR and this one should be ready to merge. And I'll add a tip about the color profile to the docs.

cbutcosk:
Some interesting discoveries here @geealbers ! I'd recently encountered some chrome-only quirks to color profiles that make more sense now, as I believe by default it forces sRGB but can use different profiles internally if configured to do so -- I wonder if PrinceXML is using that facility in their captive chromium that's not mirrored in pagedjs.

Anyway this all looks copacetic with the updated image and style, I've approved the starter PR and approving this now too.

cbutcosk (approved):

PR #1254, opened 2026-06-10
Correct logic and behavior of `FigureMedia`, external IIIF images in TOC grids

### LINKED ISSUE #1232: External IIIF Images Don't Show Up in `layout: table-of-contents` `presentation: grid`
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [x] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

When an external IIIF image is used for an object entry, it should show up in the table of contents grid layout. like this:

Image

### Actual Behavior

When an external IIIF image is used for an object entry, the card appears without an image but if you click through to the actual entry page the image appears. 

Image

The image also appears if you change the page layout to `objects-page` and remove `presentation: grid`. 

### Steps to Reproduce

1. Start a new Quire project
2. Modify `catalogue/index.md` to set `layout: table-of-contents` and `presentation: grid` 
3. Add the following entry featuring an external IIIF image to the `figures.yaml` file:
```YAML
  - id: "irises"
    iiif_image: https://media.getty.edu/iiif/image/8c255d80-7382-46db-9fa8-892c0d37247e/
    caption: "Vincent van Gogh (Dutch, 1853-1890), *Irises*, 1889. Oil on canvas. Unframed: 74.3 × 94.3 cm (29 1/4 × 37 1/8 in.), Framed [Outer Dim]: 94.9 × 114.9 × 11.4 cm (37 3/8 × 45 1/4 × 4 1/2 in.). The J. Paul Getty Museum, Los Angeles, [90.PA.20](https://www.getty.edu/art/collection/object/103JNH)."
    zoom: true
```
4. In the `objects.yaml` file add the following entry: 

```YAML
  - id: "cat-2"
    artist: Vincent Van Gogh
    title: "*Irises*"
    year: "1889"
    dimensions: 94.9 × 114.9 × 11.4 cm 
    medium: Oil on canvas
    credit: The J. Paul Getty Museum, Los Angeles
    link: https://www.getty.edu/art/collection/object/103JNH
    figure:
      - id: "irises"
  ```

5. In the `2.md` file, change the object id to "cat-2". 

### Version Numbers

[test-project]
 quire-cli 1.0.0-rc.33
 quire-11ty 1.0.0-rc.40
 starter https://github.com/thegetty/quire-starter-default@2.10.0
[System]
 quire-cli 1.0.0-rc.33
 node v22.20.0
 npm 10.9.3
 os Darwin 24.6.0

### Web Browser

_No response_

### Relevant Terminal/Shell Output

```shell

```

### Supporting Information

This bug was reported by @fullerpault in [Discussion #1186](https://github.com/thegetty/quire/discussions/1186)

cbutcosk:
This PR corrects behaviors of the `FigureMedia` class to resolve #1232 and for overall correctness:
- Refactors `annotations` and `sequences` computed properties to store factory-created annotations in a private property. The factory had been creating new class instances on every call to the computed property, leading to undefined debug behaviors and multiply-logged errors on instance creation failure.
- Eliminate unexpected behaviors for annoations by moving the execution of annotation asset handling to after the figure's core image assets have been processed. `Annotation` instances immediately produce JSON for their IIIF representation so they require asset paths populated by the image asset handling method.   
- Rename asset handling methods for clarity, eg from `processFigureSequence` to `processSequenceMedia`. This continues recent PRs clarifying when class-internal methods and properties are handling media asset data like paths to image files vs handling user-supplied figure content like captions.
- Refactors TOC grid item image component to use pre-calculated derivative media paths for applicable images, adds a test ensuring that behavior.

This branch is released for QA as `@thegetty/quire-11ty@1.0.0-rc.48`.


geealbers:
@cbutcosk, is this PR still meant to be a draft, or should I do QA on it?

cbutcosk:
@geealbers I was just waiting on resolution of the conflicts from merging the figures PR it elaborates. All set and ready for your QA.

geealbers:
@cbutcosk the only issue I'm seeing is that the 280px thumbnail size isn't large enough in many cases depending on the viewport width and the varying number of columns the TOC grid and Objects Page grid displays in. The result is a visibly pixelated/blurry image. What do you think about adding `sizes` and `srcset` to image elements on those pages so the browser will load the optimal size?

I haven't used those attributes before myself, so I don't know if there are downsides or if it would add too much complexity to the Quire templates, but if it's a viable idea I would recommend changing the thumbnail size to 360px and the static-inline-figure-image to 640px (`_plugins/figures/iiif/config.js`) and then breaking down the image display as follows.

- up to 462px width = thumbnail
- 463–699px width = static-inline-figure-image
- 700-885px width = thumbnail
- 886–1183px width = static-inline-figure-image
- 1184–1734px width = thumbnail 
- 1734px width and larger = static-inline-figure-image

This would work on both the TOC grid and Objects Page grid. And fwiw, the breakpoints for those are currently:

- 700px = 2 columns
- 1184px = 3 columns
- 1376px = 4 columns

cbutcosk:
@geealbers Hmm good point, I hadn’t accounted for the various breakpoints. What would you think of using the static inline figure image on that screen? At its largest that image will be ~590px (for a viewport 699pt wide, removing ~100pt for margin / padding) so the static inline would give at least 1x coverage for that size.

Basically responsive markup is lovely and useful but unergonomic for managing image sizing in relation to viewport widths when the component is far descended in the element tree. Reasoning about viewport sizes across column breakpoints and taking into account padding / margin is a lot when image intrinsic sizing is already hard to reason about.

So, serving a larger thumbnail gives about the same network performance (many clients will download all srcset images to evade fingerprinting) and similar rendering performance (for smaller size like portrait mobile it would give nearly 2x pixel density).

geealbers:
Sounds good @cbutcosk. I'm fine using the static inline image across the board for all the reasons you point out. Please go ahead and increase its width to 640px though. That's a better size for it when it's used inline in the text on other pages, and only a little bigger than what it's now set at.

cbutcosk:
@geealbers Great, I bumped static inline figure images to 640px and thumbnails to 320px (to keep widths even multiples for simpler intrinsic size math) and the TOC grid now uses static inlines. It looks much better at the larger edges of the breakpoints

geealbers (approved):

geealbers:
Once @bulbil has a chance to review, this is ready to merge.

bulbil (approved):
Apologies, I thought I'd responded this morning but got sidetracked. Admittedly not having gone too deep on the issues being addressed, this code all makes sense to me.

packages/11ty/_plugins/figures/figureMedia/index.js
```diff
@@ -28,6 +28,9 @@ const logger = chalkFactory('Figures:FigureMedia', 'DEBUG')
  * @property {String} printImage Optional path to an alternate image to use in print
  */
 export default class FigureMedia {
+  #annotations
+  #sequences
```
bulbil, line 40: might be worth throwing a comment here as we don't (yet) make a ton of use of private attributes and these could get lost.

@cbutcosk this brings up maybe a sensitive subject but we try to have consistent linting across our apps. I forgot to ask that in our meeting, whether there's will or whether it's even worth it to consider that. It would be kind of painful. For example, we very much require semicolons.

cbutcosk: @bulbil Good point, I updated the class with a jsdoc annotation for those properties.

Re: linting, the big question is linting alignment. quire-11ty uses an older eslint config and is passing on that preset but the config needs migrated to eslint 10+ and away from [`eslint-config-standard`](https://www.npmjs.com/package/eslint-config-standard). quire-cli has a linting configuration but it is not aligned with the quire-11ty configuration and is failing now / unused.

ETA: Agreed about the need for semicolons!

PR #1279, opened 2026-08-19
Refactor figures plugin to handle annex images (contributors, logos, etc)

cbutcosk:
This PR refactors the figures plugin to process dimensions metadata and paths about "annex" images from the quire configuration -- epub cover images, promo images, contributor images, and publisher logos. This better aligns plugin functions that share image media dimension checking and asset path calculation requirements and lifts those functions away from front-end components.

This PR is released to `npm` and github releases as `@thegetty/quire-11ty@1.0.1-rc.1`.

This code is responsive to one half of #1197. Some specific implementation / review notes:
  - These changes retain `vite`'s role in bundling these assets for the site's bundle so most of the components affected (exception: `meta` tags) still emit publication-internal paths.
  - Partially handles an issue (#1280) with `og:image` and `twitter:image` tags where assets were either nonexistent or emitted as absolute filepaths.
  - Refactors a few `package/11ty` test harness functions and bootstrap data to better represent actual values.
  - Adds tests and refactors `figures` plugin to better generalize its metadata and path functions for other types of figures, introduces the switching dispatch over `mediaType` that enables the latter half of #1197.
  - Adds an integration test ensuring that og:image tags return assets that will be served in the site bundle.
  - Adds `height` and `width` to contributor avatar images. Adding intrinsic dimensions to the publisher footer images distorted the default layout so are not included here.

PR #1285, opened 2026-08-26
Update figures plugin to handle poster images and self-hosted video media

cbutcosk:
This PR refactors the figures plugin so it manages derivatives for video and audio figures that use `poster` images and lifts the concern for poster image and media asset pathing out of the component. This is the second and final part of #1197, where all path calculations and derivative handling is now located in the figures plugin and emitted as data for direct consumption by page components. It is based on #1279 and should be reviewed after that PR.

Review notes and implementation details:
- Refactors existing figure media factory tests to better use sinon’s `fake` and `match` APIs for less test code repetition and complexity.
- Adds test paths ensuring that poster images for video and audio figures are processed, and that calculated paths for self-hosted video files are included in figure media objects. 
- Adds conditionals to `processImageMedia()` to have the factory use the figure’s `poster` value as the source for the print, static inline, etc images.
- Refactors existing figure audio, video components to use pre-calculated paths for poster and print images and self-hosted video asset paths.
- Fixes a regression from #1279 where missing image dimensions cause builds to fail.

### DIFF
--- a/packages/11ty/_plugins/figures/figureMedia/index.js
+++ b/packages/11ty/_plugins/figures/figureMedia/index.js
@@ -111,6 +111,7 @@ export default class FigureMedia {
       label,
       media_id: mediaId,
       media_type: mediaType,
+      poster,
       src,
       zoom
     } = data
@@ -147,6 +148,7 @@ export default class FigureMedia {
     this.outputDir = outputDir
     this.outputPathname = outputPathname
     this.outputFormat = format && format.output
+    this.poster = poster
     this.processImage = imageProcessor
     this.src = src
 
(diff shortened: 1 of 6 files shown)


geealbers (approved):

packages/11ty/_plugins/figures/figureMedia/index.js
```diff
@@ -379,27 +381,49 @@ export default class FigureMedia {
...
+
+      // Only handle image dimensions from media embeds with posters
+      case (this.mediaType === 'video'):
+      case (this.mediaType === 'soundcloud'):
+      case (this.mediaType === 'youtube'):
+      case (this.mediaType === 'vimeo'):
+        if (!this.poster) return
+
+        try {
+          ({ height, width } = await sharp(this.imageFilePath).metadata())
```
bulbil: we don't want to grab the dimensions straight from the `poster` here?

bulbil (approved):
One question but otherwise looks right to me.

packages/11ty/_plugins/figures/figureMedia/index.js
cbutcosk: Indeed it's true, using a getter there masks the path operations a bit too much. I clarified by having the branch now calculate this path itself (imagesDir and the rest) and then pass it to `sharp`.

PR #1290, opened 2026-08-31
Refine figure embed URL handling and fix regression for http(s) figure images

### LINKED ISSUE #1197: Transform global data (figure.src)

### LINKED ISSUE #1289: Figure images with http(s) sources generate an error with `@thegetty/quire-11ty`@`1.0.1-rc.1`
### Before proceeding, make sure there isn’t an existing issue for this bug.

- [x] I have searched the existing issues and determined this is a new bug.

### Expected Behavior

Figures with `src` properties that are URLs are supposed to be useable as-is, with the URL used in place of all browser-backed sources and a derivative only generated for the epub package.

### Actual Behavior

Per @andy-jdl in https://github.com/thegetty/quire/discussions/1287 , figure images with URLs for `src` properties generate an error:
```
[quire] ERROR Figures:FigureMedia Could not read metadata for figure cat-1: Error: Input file is missing: /Users/andyair/Desktop/code/quire/template/myproj/content/_assets/images/https:/upload.wikimedia.org/wikipedia/commons/3/3a/Cat03.jpg!
```

### Steps to Reproduce

- Clone the quire repo at `d7821666172b329a1803ba1f6ffe345cbdc677f3` or use `npm pack @thegetty/quire-11ty@1.0.1-rc.1` to download the tarball and unpack it. (NB: Manual clone steps are necessary as a workaround for #1288 ) 
- Start a default publication with the tarball contents or repo clone passed as --quire-path.
- Replace the `lange` figure block in `content/_data/figures.yaml` with: 
```
  - id: "lange"
    src: 'https://upload.wikimedia.org/wikipedia/commons/3/3a/Cat03.jpg'
    caption: "*Dorothea Lange, Resettlement Administration photographer, in California*, 1936."
    credit: "Library of Congress Prints and Photographs Division"
    alt: "Shot low and looking up at an old model car on a dirt row with a hill behind. A woman sits on the roof of the car holding a large camera and smiling."
```
- Build the publication
- Publication produces an error and fails to build.

### Version Numbers

[CLI] Command 'info' called with options { debug: true }
[test-stuff]
 quire-cli 1.0.0
 quire-11ty 1.0.1-rc.1
 starter https://github.com/thegetty/quire-starter-default@2.11.0
[System]
 quire-cli 1.0.0
 node v22.19.0
 npm 10.9.3
 os Darwin 24.6.0

### Web Browser

_No response_

### Relevant Terminal/Shell Output

```shell

```

### Supporting Information

_No response_

cbutcosk:
This PR refactors the figures plugin to normalize URL calculations for embed figures (youtube, vimeo, soundcloud) and fix a regression in http(s) figures (see #1289). It is the final part of #1197 . Details:
- Adds a test ensuring that figure embed URLs are generated on the `figureMedia` object and are properly formatted.
- Adds a test ensuring that figure images with http(s) URLs download the image URL for use in dimension calculations.
- Moves figure media embed calculations into `FigureMedia` processing loop, inserting `embedUrl` and `sourceUrl` on `derivatives.embed`. Refactors `figureMediaEmbedUrl` component to be a helper function of the figures plugin. 
- Adds a code path to `FigureMedia` that fetches http(s) URLs for use in dimensions calculation. Removes component-side figure embed calculation.


packages/11ty/_includes/components/figure/audio/element.js
```diff
@@ -39,7 +37,8 @@ export default function (eleventyConfig) {
       `
     }
   }
-  return function ({ id, mediaId, mediaType, lazyLoading }) {
-    return audioElements[mediaType]({ id, mediaId, mediaType, lazyLoading })
+  return function ({ derivatives, id, mediaId, mediaType, lazyLoading }) {
+    console.log('sup', derivatives)
```
geealbers, line 41: This should be removed, right?

geealbers (changes requested):
The logic for the display of the embed URL seems to have erroneously flipped in this PR. The URL is meant to display in the PDF and EPUB outputs (where the audio/video assets are static images) and not in the online edition. Instead, it's now showing online, but not in the PDF and EPUB.

Image

packages/11ty/_includes/components/figure/audio/element.js
cbutcosk: Yes very much so!

cbutcosk:
@geealbers Thanks for the review, well-spotted. I removed that log line and restored print caption source URLs.

packages/11ty/_plugins/figures/figureMedia/index.js
bulbil: would there be a case where the poster would be an external src url?

bulbil (approved):
Again, mini-question but otherwise looks good from my end.

packages/11ty/_plugins/figures/figureMedia/index.js
cbutcosk: Good point, and easy enough to add now that the plugin handle most all of the asset path computations. I added a test to ensure that the paths emitted are correct as well.

PR #1294, opened 2026-09-11
Remove spurious patches dir and unnecessary / inactive `fs` dependency

### LINKED ISSUE #1283: Remove `_patches` directory
The `_patches` directory is an artifact of an earlier and now-removed approach to dependency patching in `quire-11ty`. It was removed many releases ago as it got in the way of testing and did not work as intended but the patches directory remains in the root of `quire-11ty`. This issue is to remove the directory.

cbutcosk:
This PR is a small maintenance update to `quire-11ty` that removes the `_patches` directory (resolving #1283) and removes its inclusion of [`fs` in its dependencies](https://www.npmjs.com/package/fs). It also removes the CLI's `postinstall` script.


abrin (commented):
@cbutcosk I would love to be a bit more aggressive here and actually remove the `postinstall` script entirely if we could: https://github.com/thegetty/quire/blob/d7821666172b329a1803ba1f6ffe345cbdc677f3/packages/cli/scripts/postinstall.js#L24. I don't know how you feel about them, but I try and avoid this approach whenever possible

cbutcosk:
@abrin _Very_ much agreed, I try to avoid postinstall lifecycle scripts if at all possible -- removed. The script doesn't seem to _operatively_ run when I install via `npm i -g @thegetty/quire-cli` anyway.

abrin (approved):
👍

